Issue #1, opened 2026-09-11
Action Fusion renderer test fails in a default-named SoL-Pi checkout


gaoanze888:
## Summary

A fresh clone into the repository's default `SoL-Pi` directory fails the Action Fusion renderer test because the assertion searches the entire rendered output for `SoL-Pi`. Pi's built-in write renderer includes an OSC 8 file URL containing the absolute working-directory path, so the repository directory name itself triggers the assertion.

## Reproduction

```bash
git clone https://github.com/NVlabs/SoL-Pi.git
cd SoL-Pi
npm ci --ignore-scripts
npm test -- --run tests/action-fusion.test.ts
```

Observed on commit `8f8c13916c97fb54c32008e41c05421f9b69ac3b` with:

```text
Node.js v22.22.2
npm 10.9.7
```

## Actual result

```text
FAIL  tests/action-fusion.test.ts > action fusion then_run > renders a fused mutation as an English lightning savings call
AssertionError: expected 'write ...' not to contain 'SoL-Pi'

Received:
write <OSC 8 link to file:///private/tmp/solpi-repro/SoL-Pi/target.ts>target.ts...

 tests/action-fusion.test.ts:140:36
 expect(componentText(plain)).not.toContain("SoL-Pi");
```

The focused suite reports 13 passed and 1 failed. The same failure also prevents the documented `npm run check` command from completing in a default-named checkout.

## Expected result

The plain built-in renderer check should distinguish SoL-Pi's Action Fusion branding from an unrelated occurrence of `SoL-Pi` inside an absolute file URL.

For example, the assertion could target the complete branding string (`⚡ SoL-Pi · Action Fusion`) or the test could use a `cwd` whose path does not contain the repository name.

HaozheLiu-ST:
@Owen718 Please help have a check; it seems like a thing we need to address.

PR #4, opened 2026-09-11
fix: preserve file URL targets in Action Fusion

### LINKED ISSUE #2: Action Fusion skips then_run after a successful write to a file URL
## Summary

Action Fusion independently resolves the target path used for its post-mutation hash check. Its resolver does not handle `file://` URLs, although the wrapped Pi 0.84.2 `write` and `edit` tools do. As a result, a fused mutation using a valid file URL writes the requested file successfully but then skips `then_run` with `ENOENT`.

## Reproduction

On commit `8f8c13916c97fb54c32008e41c05421f9b69ac3b`, Node.js `v22.22.2`, npm `10.9.7`:

```js
import { mkdtemp, readFile } from "node:fs/promises";
import { tmpdir } from "node:os";
import { join } from "node:path";
import { pathToFileURL } from "node:url";
import { createActionFusionExtension } from "./src/sol-pi/extensions/action-fusion/index.ts";

const tools = new Map();
createActionFusionExtension({
  bashOptions: { operations: { exec: async () => ({ exitCode: 0 }) } },
})({ registerTool: (tool) => tools.set(tool.name, tool) });

const cwd = await mkdtemp(join(tmpdir(), "solpi-file-url-"));
const target = join(cwd, "target.txt");
const context = {
  cwd,
  mode: "json",
  hasUI: false,
  model: undefined,
  sessionManager: {
    getSessionId: () => "repro",
    getSessionFile: () => undefined,
  },
  ui: {},
};

try {
  await tools.get("write").execute(
    "call",
    {
      path: pathToFileURL(target).href,
      content: "written\n",
      then_run: { command: "true" },
    },
    undefined,
    undefined,
    context,
  );
} catch (error) {
  console.log(String(error));
}
console.log(await readFile(target, "utf8"));
```

Run from the repository root with:

```bash
node --experimental-strip-types repro.mjs
```

## Actual result

The mutation succeeds and `target.txt` contains `written\n`, but the tool call reports:

```text
Error: [then_run:skipped] ENOENT: no such file or directory, open '.../file:/.../target.txt'; the command was not run.
```

`src/sol-pi/extensions/action-fusion/file-queue.ts:16-20` passes the URL string to `node:path.resolve()`, producing a path containing `file:`. That path is later read by `assertUnchangedBeforeCommand()`.

Pi 0.84.2's wrapped mutation tools use `resolveToCwd()`, whose underlying path normalization converts `file://` inputs with `fileURLToPath()`. Therefore the mutation and Action Fusion's guard operate on different paths.

## Expected result

Action Fusion should apply the same path semantics as the Pi mutation tool it wraps, so the hash check targets the file that was actually written and `then_run` executes.

This likely also warrants parity tests for the other normalization handled by Pi's resolver, such as Unicode spaces and Windows shell-style drive paths, but only the `file://` case above was reproduced on this host.

Owen718:
## Summary

Fixes #2.

Decode `file://` targets with Node's `fileURLToPath()` before resolving Action Fusion's file queue and post-mutation hash guard. Pi already accepts these URLs; the guard must inspect the same file that Pi wrote or edited.

This keeps the existing queue, hash checks and command-failure behavior intact. It does not patch Pi, import private Pi APIs, or change dependencies.

## Regression coverage

Ten new tests cover file URLs and `@file://` inputs, percent-encoded spaces/`#`/`%`, real writes and edits, a real bash follow-up, serialization between URL and ordinary-path aliases, invalid URL rejection, and unchanged ordinary-path handling.

Before the runtime fix, five of the initial seven regression tests failed, including successful file mutations followed by `then_run:skipped`/`ENOENT`. The corrected implementation passes the expanded suite.

## Validation

- `npm ci --ignore-scripts`: passed; lockfile unchanged.
- `npm run check`: passed, 132 tests across 17 files, plus typecheck and package dry run.
- `npx --no-install vitest run tests/all-mechanisms.test.ts`: passed, 4 tests.
- `npm audit --audit-level=high`: passed. The unchanged lockfile currently has two moderate Vitest-related advisories; no dependency changes are included.

Tested with Node 26.0.0 and Pi 0.84.2. Based directly on `main`, independent of the renderer-test fix for #1; validation used a checkout path without the case-sensitive `SoL-Pi` substring.

### DIFF
--- a/src/sol-pi/extensions/action-fusion/file-queue.ts
+++ b/src/sol-pi/extensions/action-fusion/file-queue.ts
@@ -6,6 +6,7 @@
 import { realpath } from "node:fs/promises";
 import { homedir } from "node:os";
 import { basename, dirname, resolve } from "node:path";
+import { fileURLToPath } from "node:url";
 
 const queueTails = new Map<string, Promise<void>>();
 
@@ -14,7 +15,9 @@ function stripToolPathPrefix(filePath: string): string {
 }
 
 export function resolveToolPath(cwd: string, filePath: string): string {
-	const expanded = stripToolPathPrefix(filePath);
+	const stripped = stripToolPathPrefix(filePath);
+	// Pi accepts file URLs; the queue and hash guard must use the same target.
+	const expanded = stripped.startsWith("file://") ? fileURLToPath(stripped) : stripped;
 	if (expanded === "~") return homedir();
 	if (expanded.startsWith("~/")) return resolve(homedir(), expanded.slice(2));
 	return resolve(cwd, expanded);
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -17,6 +17,8 @@ SoL-Pi imports only public package exports:
 
 The built-in edit/write definitions capture their working directory, so SoL-Pi caches one definition per `ctx.cwd`. Its own per-file queue surrounds the built-in mutation and follow-up command. It does not nest Pi's built-in mutation queue.
 
+Action Fusion decodes `file://` targets with Node's `fileURLToPath()` before resolving the queue and hash-check path. This keeps file URLs, including percent-encoded filenames and Pi's optional `@` prefix, aligned with the file handled by the built-in mutation tool.
+
 The queue covers only fused operations registered by this SoL-Pi instance. External processes, direct built-in-tool calls outside the replacement, and unrelated extensions are not globally locked. SoL-Pi hashes the target immediately before launching `then_run` and skips the command if it observes an intervening content change.
 
 ## ObservationPack
--- a/tests/action-fusion-paths.test.ts
+++ b/tests/action-fusion-paths.test.ts
@@ -1,0 +1,154 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: MIT
+ */
+import { mkdtemp, readFile, rm, writeFile } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { join, resolve } from "node:path";
+import { pathToFileURL } from "node:url";
+import type { ExtensionAPI, ExtensionContext, ToolDefinition } from "@earendil-works/pi-coding-agent";
+import { afterEach, describe, expect, it } from "vitest";
+import { resolveToolPath } from "../src/sol-pi/extensions/action-fusion/file-queue.ts";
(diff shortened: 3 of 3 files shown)


src/sol-pi/extensions/action-fusion/file-queue.ts
```diff
@@ -14,7 +15,9 @@ function stripToolPathPrefix(filePath: string): string {
 }
 
 export function resolveToolPath(cwd: string, filePath: string): string {
-	const expanded = stripToolPathPrefix(filePath);
+	const stripped = stripToolPathPrefix(filePath);
+	// Pi accepts file URLs; the queue and hash guard must use the same target.
+	const expanded = stripped.startsWith("file://") ? fileURLToPath(stripped) : stripped;
```
QihangCao, line 20: **[P2] Malformed `file://` URLs lose the `[then_run:skipped]` marker**

`resolveToolPath()` now calls `fileURLToPath(stripped)` here, but it's evaluated as an argument to `executeMutationThenRun(...)` in `index.ts:93` / `index.ts:127` — i.e. **before** that function's try/catch even starts. If the URL is malformed (e.g. the `%2F`-encoded case your own new test covers), `fileURLToPath` throws synchronously and the error bypasses both skip-wrapping sites (`then-run.ts:97-104` around `mutate()`, and `assertUnchangedBeforeCommand`'s own catch at `then-run.ts:54-68`).

Before this PR, the same malformed input would have failed inside `mutate()` (Pi's own resolver rejecting it) and gotten properly wrapped with `[then_run:skipped]`. That marker is a tested contract elsewhere in this repo (`tests/action-fusion.test.ts:289`, `:414`), so this is a real, if narrow, regression against it — just not covered by the new test, which only asserts `.rejects.toThrow()` without checking the message.

Confirmed reproducible:
```
fileURLToPath("file:///.../bad%2Fname.txt")
→ TypeError: File URL path must not include encoded / characters
```

Suggested fix: move the `fileURLToPath` decode inside a guarded path that still produces `[then_run:skipped]` on failure — e.g. catch it in `resolveToolPath`'s caller and translate to the skipped-result shape, or do the decode lazily inside `executeMutationThenRun`'s existing try/catch instead of at the call site.

No data-loss or security impact — the existing test already confirms no mutation/command runs. This is purely an error-reporting/UX regression for one edge case.

QihangCao (commented):
Core fix is correct — it resolves #2 (file URLs no longer wrongly skip `then_run`). Found one edge-case regression in error reporting for malformed file URLs; not blocking, but worth a follow-up before merge.

PR #6, opened 2026-09-11
fix: price compaction from the actual removable prefix

### LINKED ISSUE #5: Online Context Compact can overestimate the removable prefix at message boundaries
## Summary

Online Context Compact estimates removable history as:

```ts
archiveTokens = writeTokens - systemPromptTokens - keepRecentTokens
```

However, Pi compacts only at valid message/turn boundaries. When one recent message is larger than `keepRecentTokens`, Pi may retain that whole message and summarize only a very small older prefix. OCC can therefore overestimate the removable prefix, pass its positive-savings gate, and select a compaction that is negative under its own 1,000-token memo estimate.

## Reproduction

Observed on `main` commit `22277b7e0c3c46ba1259a6687f31fe39ade421a5`, Node.js `v22.22.2`, with the pinned Pi `0.84.2`.

The following reproduces the decision using Pi's public `findCutPoint()`, `sessionEntryToContextMessages()`, and `estimateTokens()` exports:

```js
import {
  estimateTokens,
  findCutPoint,
  sessionEntryToContextMessages,
} from "@earendil-works/pi-coding-agent";
import {
  decideCompaction,
  DEFAULT_COMPACTION_ECONOMICS,
} from "./src/sol-pi/extensions/online-context-compact/economics.ts";

const usage = {
  input: 0,
  output: 0,
  cacheRead: 0,
  cacheWrite: 0,
  totalTokens: 0,
  cost: { input: 0, output: 0, cacheRead: 0, cacheWrite: 0, total: 0 },
};

const entries = [
  {
    type: "message",
    id: "old-user",
    parentId: null,
    timestamp: new Date().toISOString(),
    message: { role: "user", content: "x".repeat(100), timestamp: 1 },
  },
  {
    type: "message",
    id: "large-recent-assistant",
    parentId: "old-user",
    timestamp: new Date().toISOString(),
    message: {
      role: "assistant",
      content: [{ type: "text", text: "y".repeat(100_000) }],
      api: "test",
      provider: "test",
      model: "test",
      usage,
      stopReason: "stop",
      timestamp: 2,
    },
  },
  {
    type: "message",
    id: "abort-marker",
    parentId: "large-recent-assistant",
    timestamp: new Date(0).toISOString(),
    message: {
      role: "assistant",
      content: [],
      api: "sol-pi",
      provider: "sol-pi",
      model: "aborted",
      usage,
      stopReason: "aborted",
      timestamp: 0,
    },
  },
];

const keepRecentTokens = 20_000;
const cut = findCutPoint(entries, 0, entries.length, keepRecentTokens);
const historyEnd = cut.isSplitTurn ? cut.turnStartIndex : cut.firstKeptEntryIndex;
const summarizedEntries = [
  ...entries.slice(0, historyEnd),
  ...(cut.isSplitTurn ? entries.slice(cut.turnStartIndex, cut.firstKeptEntryIndex) : []),
];
const actualPrefixTokens = summarizedEntries
  .flatMap((entry) => sessionEntryToContextMessages(entry))
  .reduce((total, message) => total + estimateTokens(message), 0);

const writeTokens = 25_025;
const archiveTokens = writeTokens - keepRecentTokens;
const decision = decideCompaction({
  writeTokens,
  archiveTokens,
  memoTokens: 1_000,
  contextTokens: writeTokens,
  completedBoundaryRequestCounts: [1],
  remainingBoundaries: 1,
  averageContextTokenIncrement: null,
  contextWindowTokens: 30_000,
  priorCompactionCount: 0,
  carriedDebtTokens: 0,
  cacheDebtRepaymentTokens: 0,
  cacheWriteReadRatio: 12.5,
  economics: DEFAULT_COMPACTION_ECONOMICS,
});

console.log({ cut, actualPrefixTokens, archiveTokens, decision });
```

Run from the repository root:

```bash
node --experimental-strip-types repro.mjs
```

Relevant output:

```text
cut = { firstKeptEntryIndex: 1, turnStartIndex: 0, isSplitTurn: true }
actualPrefixTokens = 25
archiveTokens = 5025
decision.compact = true
decision.reason = "window_protection"
```

## Actual behavior

Pi's cut point retains the complete 25,000-token recent message and summarizes only the 25-token old prefix. OCC instead supplies `5,025` as `archiveTokens`, so its 1,000-token memo estimate reports a projected saving of `4,025` tokens and allows compaction.

Using the actual cut-point prefix with the same memo estimate gives `25 - 1,000 = -975` tokens. The same overestimate is subsequently used for cache-debt repayment and the displayed savings count.

`nativeCompactionFeasible()` prevents compaction when there are no summarizable messages, but it returns only a boolean; it does not feed the actual prefix size into the economic decision.

## Expected behavior

The positive-savings/economic decision should use an estimate of the messages that Pi's selected cut point will actually summarize, rather than total context minus a nominal retained-tail budget.

One possible direction is to extend the existing cut-point feasibility calculation to return the estimated token count of `messagesToSummarize` plus any split-turn prefix, and use that value consistently for:

- `decideCompaction().archiveTokens`;
- cache-debt repayment;
- reported token savings.

The existing economics unit test checks `archiveTokens < memoTokens`, but `archiveTokens` is supplied directly; it does not cover deriving this value from a real session cut point.

If this direction matches the intended economics, I would be happy to prepare a focused PR with the regression test.

gaoanze888:
## Summary

- estimate OCC's removable history from Pi's actual `findCutPoint()` ranges
- include a replaced previous compaction summary only when there is new history to compact
- use the cut-point-aware estimate for the economic decision, debt repayment, and savings display
- add regression coverage for oversized retained messages, repeated compactions, and no-new-prefix sessions

Fixes #5.

## Why

OCC previously used `writeTokens - systemPromptTokens - keepRecentTokens`. Pi preserves whole messages at valid cut points, so a recent message larger than the retention budget can leave only a tiny removable prefix. That could make OCC report positive savings and compact even when its own 1,000-token memo estimate implied negative savings.

For the regression fixture, the old estimate was 5,025 removable tokens while Pi's cut point exposed only 25. The new estimate follows the same history and split-turn ranges Pi summarizes, so the extension defers instead of aborting the run for an uneconomic compaction.

## Validation

```text
npm run check

Test Files  17 passed (17)
Tests       136 passed (136)
```

`npm run check` also passed TypeScript validation and `npm pack --dry-run`.

### DIFF
--- a/src/sol-pi/extensions/online-context-compact/extension.ts
+++ b/src/sol-pi/extensions/online-context-compact/extension.ts
@@ -87,13 +87,15 @@ function progressSummary(input: PlanUpdateInput, completedStepId: string): Progr
 	};
 }
 
-function compactionMessageCount(entries: readonly SessionEntry[], startIndex: number, endIndex: number): number {
-	let count = 0;
+function compactionTokenEstimate(entries: readonly SessionEntry[], startIndex: number, endIndex: number): number {
+	let tokens = 0;
 	for (let index = startIndex; index < endIndex; index++) {
 		const entry = entries[index];
-		if (entry && entry.type !== "compaction" && sessionEntryToContextMessages(entry).length > 0) count++;
+		if (!entry || entry.type === "compaction") continue;
+		const message = sessionEntryToContextMessages(entry)[0];
+		if (message) tokens += estimateTokens(message);
 	}
-	return count;
+	return tokens;
 }
 
 function branchAfterAbort(entries: readonly SessionEntry[]): SessionEntry[] {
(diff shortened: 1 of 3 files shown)


gaoanze888:

gaoanze888:
Rebased onto current `main@2b79168` (Pi 0.85.1). Exact head `b0e62bc` passes `npm run check` (19 files / 142 tests, typecheck, package dry-run), `node scripts/check-pi-compat.mjs`, the real package-loader / AgentSession integration test, and `git diff --check`. The removable-prefix pricing behavior is unchanged.

PR #8, opened 2026-09-11
perf: reuse verified reducer receipts within a session

NIU-123370:
## Problem

Repeated build/test runs can produce byte-identical diagnostic logs. Evidence-Preserving Reducer currently reuses the archived source file but still calls the reducer model for each occurrence, adding token usage and latency without new input evidence.

## Change

Reuse completed, accepted reducer responses in a per-session, in-memory LRU cache capped at 64 entries. The key covers the source hash, command, error status, configured provider/model, output limit, receipt schema, reducer instructions, and storage root.

Cache hits still verify the source archive, revalidate every quotation, and rebuild the receipt for the current tool result (including fused write/edit output). Invalid, failed, or non-shrinking reductions are not cached. Hits record a distinct `cache_hit` journal event and zero new reducer usage, avoiding duplicate accounting of the original request.

This uses existing Pi extension APIs, adds no configuration or persistent files, and does not modify Pi core. The cache is lost on process restart. Concurrent first occurrences may still issue independent requests; only completed accepted results are reused.

## Reproducible efficiency check

The new `reduces three identical logs once and charges no model usage for cache hits` test runs the reduction path with and without the cache against a mocked provider:

| Three sequential identical logs | Reducer requests |
| --- | ---: |
| Without receipt cache | 3 |
| With receipt cache | 1 |

Both cache hits retain verified evidence and report zero additional reducer tokens. This is a deterministic call-count measurement, not a paid-model benchmark or a claim about overall agent cost. Real savings depend on repeated identical logs, eviction, and provider billing.

## Validation

- `npm ci --ignore-scripts`
- `npm run check`: TypeScript, all 17 test files / 149 tests, and package dry-run passed.
- `npx vitest run tests/all-mechanisms.test.ts`: 4/4 passed.
- `git diff --check`: passed.
- `npm audit --audit-level=high`: passed; the existing lockfile reports two moderate Vitest-related findings (GHSA-82fw-gwwq-j7x9). Dependencies were not changed.

The 15 added regression cases cover repeated output, zero usage on hits, changes to source/command/status/model/provider/output limit/instructions, session/restart isolation, failed or invalid/non-shrinking receipts, archive tampering, preservation of fused results, and LRU eviction.

### DIFF
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -33,6 +33,18 @@ All persistent paths use `SessionManager.getSessionDir()` and `getSessionId()`, 
 
 The unpublished shared artifact layout is not read or migrated. Each session starts from its own `<sessionDir>/sol-pi/<sessionId>/` directory.
 
+Accepted reducer receipts are reused from a session-local, in-memory LRU cache
+of at most 64 entries. Reuse requires identical source bytes, command, error
+status, configured reducer provider/model, output limit, receipt schema, and
+reducer instructions. Every hit still verifies the source archive and validates
+the quoted evidence, then rebuilds the receipt for the current tool result.
+Hits record a `cache_hit` journal event and zero new reducer token usage; they do
+not record another provider response. Failed or rejected reductions are never
+cached. The cache adds no files and is empty after a process restart. Concurrent
+first occurrences may still make separate model calls; this cache reuses only
+completed, accepted results. Actual cost savings depend on repeated identical
+logs and the configured model's billing.
+
 ## Online Context Compact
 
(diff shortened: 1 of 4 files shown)


gaoanze888 (approved):
Reviewed the cache key, session isolation, LRU behavior, failure paths, archive revalidation, and fused-result reconstruction. I also ran `npm run check` locally: typecheck, 17 test files / 149 tests, and package dry-run all passed.

The key covers the source and reduction semantics, cache hits revalidate exact evidence against the archive, and failed/non-shrinking results are not retained. I did not find a blocking correctness issue. Nice focused optimization and clear documentation of the in-memory/session-local and concurrent-first-call trade-offs.

gaoanze888:
Integration note: approved #47 strengthens failed-result receipts so every distinct nonblank source line must be covered. Applying this current head after #47 leaves 10 cache test fixtures with partial failed-output receipts, and the combined suite fails those fixtures. The cache behavior itself is not contradicted; please rebase after #47 and update the failed-result receipt fixtures to quote all source lines before merge.

Owen718:

PR #9, opened 2026-09-11
fix: align Action Fusion Unicode-space paths with Pi

### LINKED ISSUE #7: Action Fusion skips then_run for paths containing Unicode spaces
## Summary

Action Fusion resolves the path used by its queue and post-mutation hash guard independently from Pi's wrapped `write`/`edit` tools. Pi 0.84.2 normalizes Unicode space characters in tool paths, while `resolveToolPath()` does not. With a path containing a non-breaking space, the mutation succeeds against the normalized filename but Action Fusion checks a different path, skips `then_run`, and reports an error.

## Reproduction

Reproduced on `main` commit `22277b7e0c3c46ba1259a6687f31fe39ade421a5`, Node.js `v22.22.2`, and the pinned Pi `0.84.2`.

For either fused `write` or `edit`, use a relative path containing U+00A0:

```js
const input = {
  path: "target\u00a0file.txt",
  content: "after\n",
  then_run: { command: "printf VERIFIED" },
};

await write.execute("call", input, undefined, undefined, context);
```

Pi's mutation tool writes `target file.txt` (ASCII U+0020). Action Fusion then fails while reading the unnormalized U+00A0 path:

```text
Error: [then_run:skipped] ENOENT: no such file or directory, open '.../target file.txt'; the command was not run.
```

The same result occurs with `edit`:

```text
write: ASCII-space file contains "after\n"; then_run calls = 0
edit:  ASCII-space file contains "after\n"; then_run calls = 0
```

## Root cause

Pi's built-in mutation tools resolve paths through `resolveToCwd()`, which enables `normalizeUnicodeSpaces: true`. It replaces U+00A0, U+2000–U+200A, U+202F, U+205F, and U+3000 with ASCII spaces.

Action Fusion's `resolveToolPath()` handles `@`, file URLs, and `~`, but passes other inputs directly to `node:path.resolve()`:

https://github.com/NVlabs/SoL-Pi/blob/22277b7e0c3c46ba1259a6687f31fe39ade421a5/src/sol-pi/extensions/action-fusion/file-queue.ts#L13-L23

The mutation and guard therefore address different files.

## Expected behavior

Action Fusion should use the same Unicode-space path semantics as the Pi mutation tool it wraps, so the queue/hash guard observes the file that was actually modified and `then_run` executes.

This is related to the path-parity class identified in #2, but the file-URL fix in #4 does not normalize Unicode spaces. A focused regression should cover both `write` and `edit`, assert the ASCII-space file was mutated, and assert the follow-up command ran.

I can prepare the focused fix and regression tests once the preferred parity boundary is confirmed.

gaoanze888:
## Summary

- normalize the same Unicode space characters as Pi's built-in file tools before Action Fusion resolves its queue/hash-guard path
- preserve the existing `@`, file URL, tilde, and relative/absolute path behavior
- add resolver coverage for Pi's Unicode-space set and end-to-end `write`/`edit` regression tests

Fixes #7.

## Why

Pi 0.84.2 normalizes Unicode spaces in `write` and `edit` paths. Action Fusion previously guarded the original, unnormalized path. For a path such as `target\u00A0file.txt`, Pi successfully modified `target file.txt`, but the guard then failed with `ENOENT` and skipped `then_run`.

The fix applies the same character replacement before Action Fusion strips `@` and handles file URLs, matching Pi's ordering for this path semantic. The change is intentionally limited to the Unicode-space mismatch reported in #7.

## Validation

```text
npm run check

Test Files  17 passed (17)
Tests       142 passed (142)
```

The focused tests exercise real Pi `write` and `edit` definitions, verify the normalized file contents, and confirm the follow-up command runs. TypeScript validation, package dry-run, and `git diff --check` also pass.

### DIFF
--- a/src/sol-pi/extensions/action-fusion/file-queue.ts
+++ b/src/sol-pi/extensions/action-fusion/file-queue.ts
@@ -9,13 +9,15 @@ import { basename, dirname, resolve } from "node:path";
 import { fileURLToPath } from "node:url";
 
 const queueTails = new Map<string, Promise<void>>();
+const UNICODE_SPACES = /[\u00A0\u2000-\u200A\u202F\u205F\u3000]/gu;
 
-function stripToolPathPrefix(filePath: string): string {
-	return filePath.startsWith("@") ? filePath.slice(1) : filePath;
+function normalizeToolPath(filePath: string): string {
+	const normalized = filePath.replace(UNICODE_SPACES, " ");
+	return normalized.startsWith("@") ? normalized.slice(1) : normalized;
 }
 
 export function resolveToolPath(cwd: string, filePath: string): string {
-	const stripped = stripToolPathPrefix(filePath);
+	const stripped = normalizeToolPath(filePath);
 	// Pi accepts file URLs; the queue and hash guard must use the same target.
 	const expanded = stripped.startsWith("file://") ? fileURLToPath(stripped) : stripped;
 	if (expanded === "~") return homedir();
--- a/tests/action-fusion-paths.test.ts
+++ b/tests/action-fusion-paths.test.ts
@@ -57,6 +57,14 @@ describe("Action Fusion file URL paths", () => {
 		expect(resolveToolPath(cwd, target)).toBe(target);
 	});
 
+	it.each(["\u00A0", "\u2000", "\u200A", "\u202F", "\u205F", "\u3000"])(
+		"normalizes Unicode space %s like Pi's built-in file tools",
+		(space) => {
+			const cwd = join(tmpdir(), "action-fusion-cwd");
+			expect(resolveToolPath(cwd, `target${space}file.txt`)).toBe(join(cwd, "target file.txt"));
+		},
+	);
+
 	it.each([
 		{ name: "write", prefix: "" }, { name: "edit", prefix: "" },
 		{ name: "write", prefix: "@" }, { name: "edit", prefix: "@" },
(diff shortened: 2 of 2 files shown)


gaoanze888:

gaoanze888:
A new overlapping Windows-path PR (#35) now touches the same resolver, so I checked the integration order rather than rebasing either branch blindly. Both fixes are independently valid and `npm run check` passes on each exact head, but they conflict in `file-queue.ts`.

The combined order should remain: Unicode-space normalization (#9) → optional `@` stripping → Windows drive-path normalization (#35) → `file://`/tilde/cwd resolution. This preserves both contributions and matches Pi's behavior.

@Owen718, since you recently merged #4 in this exact resolver, would you be willing to review this smaller earlier fix when convenient? If #35 lands first, I can promptly rebase #9 and resolve the single-file conflict while preserving both authors' logic.

gaoanze888:
Rebased onto current `main@2b79168` (Pi 0.85.1). Exact head `8c46b76` passes `npm run check` (19 files / 148 tests, typecheck, package dry-run), `node scripts/check-pi-compat.mjs`, the real package-loader / AgentSession integration test, and `git diff --check`. The Unicode-space path behavior is unchanged.

lyttttt3333:
@HaozheLiu-ST We independently validated this fix at `8c46b763` on native Windows Server 2022 and Ubuntu 22.04, with both Pi 0.84.2 and 0.85.1. No issues were found with the fix in the tested scenarios.

All 8 new regression tests fail against the old implementation and pass after the fix. Our independent real-file/real-shell checks also pass 92/92 after the fix in each environment, including 72 cases that previously mutated the file but skipped the follow-up command. No original test changed status.

Linux passes the full suite (148/148). Windows retains the same 3 pre-existing failures and 3 skipped tests, with no new failures from this PR (142 passed). Those existing failures keep the workflow red; they are unrelated to this change.

Validation logs: https://github.com/lyttttt3333/SoL-Pi/actions/runs/35090181603

lyttttt3333:
@gaoanze888 To reduce fragmented reviews and repeated validation work, could we consolidate these related fixes into fewer PRs?

Suggested grouping:

- **#9 + #61 + #60: Input handling and status reporting fixes** — Unicode-space paths, reducer route whitespace normalization, and displaying Action Fusion savings only after success.
- **#6 + #13 + #62: Context management and cost accounting fixes** — removable-history estimation, branch-scoped ObservationPack counters, and cumulative cache debt.
- **Keep #59 separate** — it addresses a potentially blocking llama.cpp compatibility issue and benefits from independent validation and rollback.

Please retain separate commits, linked issues, and regression tests for each fix, especially #13, which is logically independent of the compaction accounting changes. The combined versions should also be validated before merging.

We have independently validated #9 and #61 so far; this grouping proposal does not imply approval of the other fixes.

The goal is to reduce maintenance overhead while preserving clear attribution and test coverage. As a general contribution guideline, we would prefer related small fixes submitted around the same time to be grouped where practical, with urgent or blocking fixes kept separate.

lyttttt3333:
@HaozheLiu-ST We have independently validated this PR on the supported Pi versions we checked (0.84.2 and 0.85.1), across Windows and Linux. The change is small and focused on aligning Action Fusion path handling with Pi for Unicode-space paths, and our validation did not find regressions introduced by this PR.

From the validation side, this looks ready to merge. The only merge-order note is to preserve the resolver order if combined with #35: Unicode-space normalization -> optional @ stripping -> Windows drive-path normalization -> file URL / tilde / cwd resolution.

HaozheLiu-ST (approved):
LGTM. Thanks @lyttttt3333 for the independent Windows/Linux validation. Combined checks with #35 and #61 pass, including four additional path-composition regressions. The agreed normalization order will be preserved when integrating #35.

PR #13, opened 2026-09-11
fix: scope Observation Pack send counts to session branches

### LINKED ISSUE #12: Observation Pack carries send counts across session tree branches
## Summary

Observation Pack keeps per-observation send counts in an extension-instance `Map`, but does not reset or rebuild that state after `session_tree`. After navigating to an earlier point and creating a new branch, an observation can therefore be replaced immediately even though it has not received its first two full sends on the new branch.

This conflicts with the branch-local reconstruction already performed by `priorAssistantCounts`: the projected branch history says the observation has zero subsequent assistant requests, but the stale in-memory count from the abandoned branch wins.

## Reproduction

Observed on `main` commit `22277b7e0c3c46ba1259a6687f31fe39ade421a5` with the pinned Pi `0.84.2`.

A minimal lifecycle sequence is:

```ts
const body = largeToolResultOverThreshold();
const message = toolResult(body);

// Original branch.
await emitContext([message]); // full
await emitContext([message]); // full
await emitContext([message]); // placeholder

// Pi emits this after moving the active leaf to an earlier branch point.
await emit("session_tree", {
  type: "session_tree",
  oldLeafId: "old-branch",
  newLeafId: "branch-point",
  summaryEntry: undefined,
});

// First request on the new branch.
await emitContext([message]);
```

Actual sequence:

```text
original branch request 1: full
original branch request 2: full
original branch request 3: placeholder
new branch request 1:      placeholder
```

A fresh extension instance projected over the same new-branch history sends the result in full, confirming that the difference is stale process state rather than the branch contents.

The same behavior was reproduced through Pi 0.84.2's real `ExtensionRunner` with all four SoL-Pi mechanisms registered:

```text
all-enabled-pre-tree-1 full
all-enabled-pre-tree-2 full
all-enabled-pre-tree-3 placeholder
all-enabled-after-tree placeholder
all-enabled-fresh-baseline full
```

## Root cause

`sentCounts` is scoped to the extension instance and keyed only by session runtime root plus observation ID:

https://github.com/NVlabs/SoL-Pi/blob/22277b7e0c3c46ba1259a6687f31fe39ade421a5/src/sol-pi/extensions/observation-pack/index.ts#L52-L54

https://github.com/NVlabs/SoL-Pi/blob/22277b7e0c3c46ba1259a6687f31fe39ade421a5/src/sol-pi/extensions/observation-pack/index.ts#L155-L163

The extension derives a branch-local fallback from assistant messages following each result, but `sentCounts.get(key) ?? priorAssistantCounts[index]` always prefers the stale count once one exists. Observation Pack does not handle `session_tree`, although Pi updates the active branch and then emits that event before the next provider request.

## Impact

A model working from a newly-created branch may receive only the excerpt placeholder for a historical large tool result on its first request, rather than the two full sends promised by Observation Pack. It can recover through `obs_recall`, but that requires an avoidable extra tool round-trip and may omit context needed for the branch's first decision.

## Expected behavior

Send-count state should follow the active session branch. After `session_tree`, the next projection should derive counts from the newly active branch history rather than from the abandoned branch.

A focused fix could invalidate the affected session's in-memory send counts on `session_tree` and let the existing `priorAssistantCounts` reconstruction establish the branch-local baseline. Regression coverage should compare navigation behavior with a fresh extension instance over the same branch.

I would be happy to prepare the focused lifecycle fix and regression test if this branch-local interpretation matches the intended behavior.

gaoanze888:
## Summary

- scope Observation Pack's transient send counters to the session-tree leaf where a request occurs
- inherit counters only from the active branch ancestry, keeping sibling branches independent
- combine transient retry counts with assistant-history reconstruction and cap stored values after placeholder transition
- clear the current session's transient counters and ledger handle on shutdown
- add branch, ancestry, session-isolation, and shutdown regressions

Fixes #12.

## Why

The previous counter key contained only the session runtime root and observation ID. After navigating to an earlier tree point, a new sibling branch therefore inherited the abandoned branch's count and received a placeholder on its first request.

The fix attaches transient request counts to immutable session leaf IDs. A descendant inherits counts from its active ancestors, while a sibling does not see counts recorded only on the other branch. Persisted assistant history remains the baseline after restart, and in-memory state only supplements requests that have not yet produced an assistant entry.

This avoids a global reset: returning from branch B to branch A preserves A's request count instead of sending the full observation again.

## Validation

```text
npm run check

Test Files  17 passed (17)
Tests       136 passed (136)
```

Focused tests also cover sibling A→B→A navigation, ancestor-to-descendant inheritance, cross-session isolation, and session shutdown cleanup. TypeScript, package dry-run, and `git diff --check` pass.

### DIFF
--- a/src/sol-pi/extensions/observation-pack/index.ts
+++ b/src/sol-pi/extensions/observation-pack/index.ts
@@ -50,7 +50,7 @@ const RECALL_LIMITS = {
 
 export function createObservationPackExtension(): ExtensionFactory {
 	return (pi: ExtensionAPI) => {
-		const sentCounts = new Map<string, number>();
+		const sentCounts = new Map<string, Map<string, Map<string, number>>>();
 		const ledgers = new Map<string, Ledger>();
 		const ledgerFor = (ctx: ExtensionContext): Ledger => {
 			const root = runtimeRoot(ctx);
@@ -61,6 +61,12 @@ export function createObservationPackExtension(): ExtensionFactory {
 			}
 			return ledger;
 		};
+
+		pi.on("session_shutdown", (_event, ctx) => {
+			const root = runtimeRoot(ctx);
+			sentCounts.delete(root);
+			ledgers.delete(root);
+		});
 
 		pi.registerTool({
 			name: "obs_recall",
@@ -137,6 +143,19 @@ export function createObservationPackExtension(): ExtensionFactory {
 		pi.on("context", async (event, ctx: ExtensionContext) => {
 			const projected = [...event.messages];
 			const root = runtimeRoot(ctx);
+			const branchIds = ["<root>", ...ctx.sessionManager.getBranch().map((entry) => entry.id)];
+			const leafId = ctx.sessionManager.getLeafId() ?? "<root>";
+			let sessionCounts = sentCounts.get(root);
+			if (!sessionCounts) {
+				sessionCounts = new Map();
+				sentCounts.set(root, sessionCounts);
+			}
+			const activeCounts = new Map<string, number>();
+			for (const id of branchIds) {
+				for (const [observationId, count] of sessionCounts.get(id) ?? []) {
+					activeCounts.set(observationId, Math.max(activeCounts.get(observationId) ?? 0, count));
+				}
+			}
 			// How many provider requests each message has already been part of,
 			// counted by the assistant messages that follow it.
 			const priorAssistantCounts = new Array<number>(event.messages.length);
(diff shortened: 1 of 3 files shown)


gaoanze888:

gaoanze888:
Rebased onto current `main@2b79168` (Pi 0.85.1). Exact head `c4c19d6` passes `npm run check` (19 files / 142 tests, typecheck, package dry-run), `node scripts/check-pi-compat.mjs`, the real package-loader / AgentSession integration test, and `git diff --check`. The branch-scoped send-count behavior is unchanged.

PR #15, opened 2026-09-11
fix: declare the tested minimum Pi version in peer dependencies

### LINKED ISSUE #11: Peer dependency range accepts Pi versions that cannot load SoL-Pi
## Summary

SoL-Pi declares all Pi packages as `"*"` peer dependencies, but the extension does not load with older versions that satisfy that range. In an isolated install with Pi 0.79.10, the compatibility check and package import fail because `@earendil-works/pi-ai/compat` was not exported yet.

The compatibility documentation identifies Pi 0.81.1 as the exercised baseline, so the peer range should communicate that minimum instead of allowing npm to accept incompatible installations.

## Reproduction

Observed from `main` commit `22277b7e0c3c46ba1259a6687f31fe39ade421a5` with Node.js `v22.22.2`.

1. Copy the repository to an isolated directory.
2. Set these development packages to `0.79.10`:

```json
{
  "@earendil-works/pi-agent-core": "0.79.10",
  "@earendil-works/pi-ai": "0.79.10",
  "@earendil-works/pi-coding-agent": "0.79.10",
  "@earendil-works/pi-tui": "0.79.10"
}
```

3. Install and run:

```bash
npm install --ignore-scripts --legacy-peer-deps
node scripts/check-pi-compat.mjs
```

Actual result:

```text
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './compat' is not defined by "exports" in .../@earendil-works/pi-ai/package.json
```

Importing `src/sol-pi/index.ts` fails for the same reason because the top-level entry statically reaches the reducer provider module, even when Evidence-Preserving Reducer would be disabled by configuration. Pi 0.79.10 also predates public APIs used by Online Context Compact, including `agent_settled` and `sessionEntryToContextMessages`.

For comparison, the same isolated procedure with all four Pi packages at `0.81.1` passes:

```text
node scripts/check-pi-compat.mjs  # exit 0
npm run typecheck                 # exit 0
import('./src/sol-pi/index.ts')    # load ok
```

## Actual behavior

`package.json` declares:

```json
"@earendil-works/pi-agent-core": "*",
"@earendil-works/pi-ai": "*",
"@earendil-works/pi-coding-agent": "*",
"@earendil-works/pi-tui": "*"
```

Package managers therefore consider Pi 0.79.10 compatible, although SoL-Pi cannot load with it.

## Expected behavior

The peer dependency contract should express the tested minimum Pi version documented in `docs/compatibility.md` (currently 0.81.1), consistently across the Pi packages that must share a release line.

A focused regression could validate the lower-bound installation/import in CI or add a package-metadata assertion so the declared range cannot drift below the documented compatibility baseline.

I would be happy to prepare the package metadata and compatibility-check update if this minimum-version policy matches the project's intent.

dajiaohuang:
## Summary

Fixes #11.

The four Pi packages were declared as `"*"`, so npm accepted Pi 0.79.10 even though SoL-Pi cannot load against it. That release predates the `@earendil-works/pi-ai/compat` export and the `agent_settled` and `sessionEntryToContextMessages` APIs SoL-Pi imports.

The declared floor is now 0.81.1, the lowest release `docs/compatibility.md` records as exercised, and the lockfile root mirrors it. `typebox` is unchanged, and no Pi package, dependency, or runtime behavior changes.

## Regression coverage

Three assertions in `tests/package.test.ts` keep the declared contract aligned:

- every Pi peer range is identical and not `"*"`, and its floor equals the floor declared in `docs/compatibility.md`;
- the lockfile root peers match the manifest;
- the declared floor is above 0.79.10.

Each was checked against a deliberate regression. Reverting the manifest ranges to `"*"` fails all three; reverting only the lockfile entry fails the lockfile assertion, which is the drift this change also fixes.

## Validation

- Pi 0.79.10, isolated install: on the unmodified base, `node scripts/check-pi-compat.mjs` and `import('./src/sol-pi/index.ts')` both fail with `ERR_PACKAGE_PATH_NOT_EXPORTED` for `@earendil-works/pi-ai/compat`. Pi 0.81.1 passes both. `*` accepts 0.79.10; `>=0.81.1` rejects it and still accepts 0.81.1 and 0.84.2.
- `npm run typecheck`: passed.
- `npx vitest run tests/package.test.ts -t "declared Pi compatibility"`: passed, 3 tests.
- Full suite: 132 passed, 5 failed. The same 5 fail on an unmodified base checkout — two `npm pack` cases that need a POSIX `npm` shim, two symlink cases, and one reducer case — so none are regressions. `npm run check` therefore does not complete on this Windows host; the pack cases are expected to pass in CI.
- `npm ci --ignore-scripts --dry-run`: passed.
- `npm audit --audit-level=high`: passed; the two moderate Vitest advisories are unchanged.

Tested with Node 24.14.1 and Pi 0.84.2.

### DIFF
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -1,6 +1,6 @@
 # Pi Compatibility
 
-SoL-Pi is developed and tested against `@earendil-works/pi-coding-agent` 0.84.2. Its public API surface is also type-checked and exercised against Pi 0.81.1, the base used by the original Pi fork. The runtime range is deliberately expressed as a peer dependency because Pi owns installation and upgrade of its packages.
+SoL-Pi is developed and tested against `@earendil-works/pi-coding-agent` 0.84.2. Its public API surface is also type-checked and exercised against Pi 0.81.1, the base used by the original Pi fork. The runtime range is deliberately expressed as a peer dependency because Pi owns installation and upgrade of its packages; the declared floor is 0.81.1, the lowest exercised release, because earlier Pi versions predate the `@earendil-works/pi-ai/compat` export and the `agent_settled` and `sessionEntryToContextMessages` APIs SoL-Pi imports.
 
 SoL-Pi imports only public package exports:
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,10 +22,10 @@
 				"node": ">=22.19.0"
 			},
 			"peerDependencies": {
-				"@earendil-works/pi-agent-core": "*",
-				"@earendil-works/pi-ai": "*",
-				"@earendil-works/pi-coding-agent": "*",
-				"@earendil-works/pi-tui": "*",
+				"@earendil-works/pi-agent-core": ">=0.81.1",
+				"@earendil-works/pi-ai": ">=0.81.1",
+				"@earendil-works/pi-coding-agent": ">=0.81.1",
+				"@earendil-works/pi-tui": ">=0.81.1",
 				"typebox": "*"
 			}
 		},
--- a/package.json
+++ b/package.json
@@ -34,10 +34,10 @@
 		]
 	},
 	"peerDependencies": {
-		"@earendil-works/pi-agent-core": "*",
-		"@earendil-works/pi-ai": "*",
-		"@earendil-works/pi-coding-agent": "*",
-		"@earendil-works/pi-tui": "*",
+		"@earendil-works/pi-agent-core": ">=0.81.1",
+		"@earendil-works/pi-ai": ">=0.81.1",
+		"@earendil-works/pi-coding-agent": ">=0.81.1",
+		"@earendil-works/pi-tui": ">=0.81.1",
 		"typebox": "*"
 	},
 	"devDependencies": {
--- a/tests/package.test.ts
+++ b/tests/package.test.ts
@@ -11,6 +11,39 @@ interface PackReport {
 	files: Array<{ path: string }>;
 }
 
+interface PackageManifest {
+	peerDependencies: Record<string, string>;
+}
+
+const piPackages = [
+	"@earendil-works/pi-agent-core",
+	"@earendil-works/pi-ai",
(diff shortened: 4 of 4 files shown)


gaoanze888 (approved):
Reviewed against the reproduction in #11. The metadata change is focused: the four shared-release Pi peers now use the documented, exercised 0.81.1 floor, with the lockfile and compatibility docs kept in sync. I ran `npm run check` locally: typecheck, 17 test files / 137 tests, and package dry-run all passed.

The added package-contract tests explain most of the diff and cover the important drift modes (wildcard ranges, package/lockfile mismatch, and the known-incompatible 0.79.10 boundary). I did not find a blocking correctness issue. Thanks for turning the issue into a well-scoped compatibility fix.

gaoanze888 (changes requested):
Re-reviewed exact head `63e73e80a5e1ddc3b92d0f813515730d5ec228d9` against current `main@2b79168`. My earlier approval is no longer current after #63: this branch is now conflicting, rewrites the new compatibility evidence, and declares `>=0.81.1` even though the current full-suite evidence explicitly does **not** establish 0.81.1 runtime compatibility (the 0.81.1 run has two Online Context Compact integration failures). The unbounded range also claims untested future Pi releases. Please rebase and replace the range with a bounded range supported by an actual lower-bound runtime matrix; current repository evidence supports 0.84.2 and 0.85.1, not 0.81.1. This also now overlaps #56, so the two should be consolidated rather than merged independently.

dajiaohuang:
Thanks for the re-review. I confirmed that #56 is still open and covers issue #11 with the same peer-range change, while merged #63 contains the current 0.84.2/0.85.1 compatibility evidence. To avoid duplicate or conflicting PRs, I’m leaving #15 unchanged rather than rebasing independently; maintainer consolidation can decide which path should carry the bounded range. No merge was performed.

Owen718:

PR #14, opened 2026-09-11
Add literal search for archived observations

aalvsz:
## Summary

- Add optional literal UTF-8 search to `obs_recall` for archived large tool results.
- Return exact byte spans, exclusive ends, LF-based line numbers, bounded UTF-8-safe source context, and continuation metadata.
- Preserve the existing paged-recall path when `query` is omitted.
- Add validation for malformed Unicode, query size, output limits, cancellation, file guards, overlaps, and continuation.
- Add a reproducible local comparison script.

## Validation

- `npm run check` — 17 test files, 142 tests, typecheck, and package dry run passed.
- `npm audit --audit-level=high` — passed at the requested severity; two pre-existing moderate Vitest advisories remain.
- `node scripts/check-pi-compat.mjs` — passed.
- `npx --no-install vitest run tests/all-mechanisms.test.ts` — 4 tests passed.
- Independent full-buffer oracle — 23 cases / 57 search calls passed.

## Real-output benchmark

Five real `rg` commands from the repository produced 30–166 KiB archived outputs. In a 50-case targeted benchmark (eight observed literals with at most 20 occurrences plus two misses per artifact):

| metric | paged recall | literal search | change |
|---|---:|---:|---:|
| tool calls | 330 | 53 | 83.94% fewer |
| returned bytes | 4,900,500 | 204,085 | 95.84% fewer |

All 50 cases were source-oracle verified, including offsets, line numbers, contexts, and continuation. A separate dense stress run is retained: queries matching nearly every line required more output than paging, which documents the feature's workload boundary.

These are retrieval measurements over real command output. They do not claim improved model-answer accuracy, provider-token usage, or billed cost. Timings are descriptive local wall times.

PR #17, opened 2026-09-11
Add live trajectory inspector

aalvsz:
Adds an opt-in live trajectory widget and a metadata-only session log so users can follow model requests, tools, and compaction as an agent runs. `/trajectory` toggles the view. Prompts, assistant text, tool arguments, and tool output are omitted.

Validation on Pi 0.84.2:
- 149 automated tests pass, plus typecheck and package validation.
- 50 matched cases (100 real SDK sessions, 300 tool executions): identical provider-request hashes and expected outputs in both arms.
- Mean local elapsed time: 7.81 ms disabled, 12.70 ms enabled. Paired overhead: mean +4.89 ms, p95 +12.34 ms. This is observability overhead, not a speed or accuracy improvement.
- Real interactive CLI display/toggling and a live-model file-read smoke passed. Long-run, resume, HTTP 503 retry, cancellation, and native compaction checks passed through Pi's runtime.

See the [method and results](https://github.com/aalvsz/SoL-Pi/blob/trajectory-inspector/docs/benchmarks/trajectory-inspector.md) and [all 50 measured pairs](https://github.com/aalvsz/SoL-Pi/blob/trajectory-inspector/docs/benchmarks/trajectory-review-50.csv). The reproducible benchmark uses a deterministic local HTTP provider; the live-model check is a separate single smoke test.


gaoanze888 (changes requested):
Thanks for the unusually thorough benchmark and lifecycle validation. I ran `npm run check` locally (18 files / 145 tests, typecheck, and package dry-run all passed), then added adversarial lifecycle probes. Two correctness gaps remain before the trajectory can be treated as authoritative:

1. **A retryable HTTP response finalizes the whole logical request too early.** `after_provider_response` updates and clears the single active request on the first callback. Pi providers can invoke the response callback for a retryable 429/5xx and then succeed on a later attempt. With the event sequence `before_provider_request → HTTP 429 → HTTP 200`, the durable request span ends permanently as `error / HTTP 429`; the 200 has no active span to update. Please retain the logical request across retryable attempts or model attempts separately, and add a `429 → 200` regression.

2. **Failed/cancelled compactions remain `running` forever.** The compaction span is closed only by `session_compact`, but Pi does not emit that extension event when native compaction throws or is cancelled. A later `session_before_compact` overwrites the reference, leaving the earlier durable record unmatched. I reproduced `session_before_compact → agent_settled → shutdown` with no update for the running span. Please choose a representation that does not promise a completion hook the public extension API cannot provide (for example an attempt/info record, or an explicit fallback closure), and cover failure/cancellation plus a later successful compaction.

These are important because an observability feature that reports a recovered request as failed or leaves completed attempts running can send debugging in the wrong direction. The concurrent tool-call pairing and the normal transport-error fallback looked correct in my review.

A separate product-level question for maintainers: the widget tail is bounded, but `events.jsonl` has no retention/rotation policy. The submitted benchmark measures about 8.8 KiB for very short sessions, so long-running trajectories should have an explicit storage bound or documented retention policy before this is recommended broadly.

aalvsz:
Addressed both lifecycle findings:

- Logical requests now remain active across response callbacks and finish on assistant completion, with turn-end as a fallback. The last HTTP status is retained. Added 429 → 200 and 503 → 200 regressions, including durable-log assertions.
- Compaction starts are informational attempts, followed by a separate success record only when Pi emits `session_compact`. Missing failure/cancellation callbacks can no longer leave a running span or be mistaken for success. Added incomplete-attempt/shutdown regressions and real SDK checks for native failure and cancellation, each followed by a successful compaction.
- Documented retention explicitly: the widget is bounded, but the ledger has no automatic rotation or size cap. Hiding the widget does not stop logging; recording can be disabled for subsequent invocations, and session logs can be archived/deleted after stopping the session.

Validation: `npm run check` passes (18 files / 149 tests, typecheck, package dry-run). All seven real SDK lifecycle scenarios pass, including the two new compaction paths.

gaoanze888 (approved):
Re-reviewed the follow-up at `f75131d5d`. Both requested lifecycle fixes are addressed: request spans remain open through retry-related response callbacks and close on assistant completion with a turn-end fallback; compaction starts are now informational attempts and only `session_compact` emits a separate success record, so failure/cancellation cannot leave a running span. `npm run check` is green (18 files / 149 tests, typecheck, package dry-run), and the retention limitation is now explicit in the docs. One wording nuance: provider callback granularity is provider-dependent (the bundled OpenAI/Anthropic retry helper may expose only the final response), so “individual response attempts” should not be read as universal; this is non-blocking for the corrected lifecycle behavior.

gaoanze888:
Merge-order note: this head has a documentation conflict with approved #14 in `docs/configuration.md`; the mechanisms are otherwise complementary. Please rebase whichever lands second and retain both the trajectory-inspector configuration and #14’s archived-observation literal-search documentation.

aalvsz:
Thanks — confirmed that #14 is still open and that `docs/configuration.md` is the only overlap. I’ll rebase #17 onto updated `main` after #14 lands, preserving both the trajectory-inspector documentation and the archived-observation literal-search documentation. No implementation changes are needed for this merge-order note.

aalvsz:
Following up on the merge-order note: I rebased `trajectory-inspector` onto #14’s current head `b073a42` and force-pushed the result as `3a12a63`. I resolved `docs/configuration.md` by retaining both the archived-observation literal-search documentation and the Trajectory Inspector lifecycle/retention documentation. The trajectory commits remain equivalent under `git range-diff`.

Validation on the rebased branch: `npm run check` (18 files / 157 tests), `node scripts/check-pi-compat.mjs`, `npx vitest run tests/all-mechanisms.test.ts`, `npm audit --audit-level=high`, and `git diff --check` all pass. The branch intentionally carries #14’s ancestry until #14 lands, after which the PR diff will reduce to the trajectory contribution.

Owen718:

PR #23, opened 2026-09-12
fix: report Action Fusion reachability and reducer minBytes fallbacks

### LINKED ISSUE #22: Mechanism reachability under a replaced tool surface: only Action Fusion goes inert (correction to #20/#21, plus two observability gaps)
Follow-up to my comments on #20 and #21. **I need to correct one thing I said there**, and the correction makes this a
narrower and more actionable report than "code mode breaks everything".

## What I claimed and what is actually true

In #20/#21 I attributed three mechanisms to a single cause ("another extension owns the tool surface, so SoL-Pi is
unreachable"). Measured properly, that is only true for **one** of the four mechanisms. The other three hook event
pipelines (`tool_result`, context projection) rather than a named tool, and they work fine when the tool surface is
taken over by a flat-tool harness (pi-fabric, code mode: one `fabric_exec` tool, model calls `pi.edit(...)` /
`pi.bash(...)` inside a sandboxed program).

## Setup

- Pi 0.85.1, SoL-Pi 0.1.0 @ `d7ecfc0`, macOS Apple Silicon, `deepseek-v4.1-flash` @ high thinking
- Isolated `PI_CODING_AGENT_DIR` / session dir, one mechanism enabled per run
- Same two profiles: **native** (SoL-Pi + a normal extension set) and **flat-tool** (same set incl. pi-fabric)
- Workloads chosen to clear each mechanism's own gate (e.g. `make` emitting ~17 KB with failure lines, full content
  returned unfiltered)

| Mechanism | Keys on | Native | Flat-tool harness |
|---|---|---|---|
| Action Fusion | replaces `edit`/`write`, expects `then_run` in their args | ✅ 5/5 edits fused, `[then_run:succeeded]` ×5 | ❌ **inert** — 36 tool calls, `then_run` present 0 times, no marker |
| ObservationPack | context projection over any tool message ≥ 10 KB (`observation.ts:13`, `THRESHOLD_BYTES = 10 * 1024`) | ✅ ledger, `read` 43 765 B → 10 942 tokens | ✅ **works** — `{"event":"full","tool":"fabric_exec","originalBytes":17747,"originalTokens":4437}` |
| Evidence-Preserving Reducer | `tool_result` handler, `event.toolName === "bash"` or `edit`/`write` + `then_run` (`candidate.ts:65-79`) | ✅ receipt in `details` + archive object | ✅ **works** — receipt + archive object written from a flat-tool run |
| Online Context Compact | boundary/compaction pipeline + `update_plan` | ✅ state entries per request | ✅ **works** — `sol-pi-online-context-state-v1` entries, `requestCount`/`lastContextTokens` tracked |

## The actual failure mode for Action Fusion

It is not "the fused tool is unreachable". Under a flat-tool harness the model never *calls* the native `edit` tool — it
calls a harness tool whose edit surface has no `then_run` parameter. So the fused `edit` may well still be the registered
tool; the model simply has no way to supply the argument that Action Fusion exists to consume, and the bare-return path
(`thenRun === undefined`) produces Pi's plain success message with no marker. Indistinguishable from a regression until
you check which tool the model actually invoked.

So the two problems reported in #20 and #21 are the same shape: a mechanism whose entry condition is *a parameter on a
specific tool* silently disappears whenever the model stops calling that tool directly.

## Two observability gaps I hit while measuring this

1. **`minBytes` declines without a journal entry.** `evidence-preserving-reducer/index.ts:67` returns `undefined`
   silently when the recovered body is under 4 096 bytes, while the neighbouring gates (`source-over-max-chars`,
   `likely-secret`) do journal a fallback. I lost a full test cycle to this: a passing `npm test` emits 262 bytes
   (vitest is quiet on success), so a perfectly qualifying command produced no artifact and no explanation. One more
   `journal("fallback", { reason: "source-under-min-bytes", bytes })` would make that self-explaining.
2. **No load-time reachability check.** Both silent failures in this thread would have been one warning line at load:
   "Action Fusion is enabled but the tool the model calls does not expose `then_run`" / "Reducer is enabled but no tool
   in this session reports as `bash`". Right now the only symptom is an absence.

Happy to test a patch for either; the harness above (two profiles, per-mechanism config, forced workloads) reproduces
both arms in a couple of minutes.

KayanoLiam:
Fixes #22's diagnostic/observability gaps; does not make Action Fusion execute through `fabric_exec`.

## Changes

- Journal short diagnostic-body declines as `fallback` / `source-under-min-bytes`, with byte count and threshold.
- At each `before_provider_request`, check the active/configured tool surface and persist branch-deduplicated `sol-pi-reachability-v1` diagnostics (not LLM context), plus TUI warnings.
- Action Fusion requires an active SoL-Pi fused edit/write registration, identified by its registered parameter object and name—not an arbitrary tool's `then_run` field.
- Reducer accepts configured bash or an active SoL-Pi fused edit/write path, matching its candidate contract.
- No provisional session-start warning. After tree navigation, the next request re-appends a warning if absent from that branch. Missing inspection APIs skip diagnostics.

Review follow-up: `bdcd78de64197e2b681f26383f5426948ad1578f`. See the review-response comment for per-item details and evidence.

## Local verification

Pinned Pi 0.84.2:
- `npx vitest run tests/reachability.test.ts tests/action-fusion.test.ts tests/evidence-preserving-reducer.test.ts`: PASS, 49 tests.
- `npm run check`: PASS, typecheck + 157 tests / 19 files + pack dry-run.
- `node scripts/check-pi-compat.mjs`: PASS (API smoke check).
- `git diff --check` / `git diff --cached --check`: PASS. No separate formatter/linter command is configured.

Additional isolated Pi 0.81.1 run:
- Same targeted tests: PASS, 49 tests.
- `npm run check`: **FAIL**, typecheck passes, 155 tests pass, 2 existing automatic-compaction AgentSession tests fail at line 157 (call counts 4/3 and 6/5).
- Those exact failures also reproduce on previous PR head `17df74c` and upstream base `d7ecfc0` with the same dependencies. They remain unresolved, not dismissed as flaky.
- Separate `node scripts/check-pi-compat.mjs` and `npm pack --dry-run`: PASS.

GitHub CI: no PR checks reported and no PR test workflow configured; these are local results only.

### DIFF
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -8,7 +8,8 @@ SoL-Pi imports only public package exports:
 - `createWriteToolDefinition`
 - `createBashToolDefinition`
 - extension types and `ExtensionAPI.registerTool`
-- `context`, `before_provider_request`, `tool_result`, `turn_end`, `agent_settled`, and `session_before_tree` extension events
+- `ExtensionAPI.getActiveTools()` and `ExtensionAPI.getAllTools()` when present, used only for load-time reachability diagnostics
+- `context`, `before_provider_request`, `before_agent_start`, `tool_result`, `turn_end`, `agent_settled`, and `session_before_tree` extension events
 - native compaction events, `ExtensionContext.getContextUsage()`, and `ExtensionContext.compact()`
 - `ExtensionContext.model` and `ExtensionContext.modelRegistry`
 - the public session-manager methods exposed through `ExtensionContext`
@@ -21,13 +22,15 @@ Action Fusion decodes `file://` targets with Node's `fileURLToPath()` before res
 
 The queue covers only fused operations registered by this SoL-Pi instance. External processes, direct built-in-tool calls outside the replacement, and unrelated extensions are not globally locked. SoL-Pi hashes the target immediately before launching `then_run` and skips the command if it observes an intervening content change.
 
+Action Fusion's entry condition is the `then_run` parameter on the tool the model actually calls. After registration, and again on the first `before_agent_start`, SoL-Pi inspects `getActiveTools()` / `getAllTools()` when those methods exist. If no model-visible tool exposes `then_run`, it appends a `sol-pi-reachability-v1` session entry (not LLM context) and shows a TUI warning. A Pi build without those inspection APIs skips the diagnostic rather than failing to load. Code-mode and other flat-tool harnesses that already combine an edit and its follow-up command inside one model turn are outside Action Fusion's contract; the diagnostic exists so that inertness is not mistaken for a regression.
+
 ## ObservationPack
 
 ObservationPack changes only the messages projected through the public `context` event. Stored session history remains intact. Original bytes and the JSONL ledger live under the session-derived SoL-Pi directory.
 
 ## Evidence-Preserving Reducer
 
-The reducer handles public `tool_result` events and resolves the configured reducer provider/model through Pi's model registry before calling `ExtensionContext.modelRegistry.complete()` when available. For the Pi 0.81.1 fork, which exposes no registry `complete()` method, it resolves authentication for that reducer model through `getApiKeyAndHeaders()` and calls the shared `@earendil-works/pi-ai/compat` completion API. The reducer preserves the original result whenever the configured reducer model is unavailable or eligibility, model-call, schema, source-hash, exact-quote, size, or likely-secret checks fail.
+The reducer handles public `tool_result` events and resolves the configured reducer provider/model through Pi's model registry before calling `ExtensionContext.modelRegistry.complete()` when available. For the Pi 0.81.1 fork, which exposes no registry `complete()` method, it resolves authentication for that reducer model through `getApiKeyAndHeaders()` and calls the shared `@earendil-works/pi-ai/compat` completion API. The reducer preserves the original result whenever the configured reducer model is unavailable or eligibility, model-call, schema, source-hash, exact-quote, size, or likely-secret checks fail. A recovered body under `minBytes` journals `fallback` with `reason: "source-under-min-bytes"` so a quiet passing test run is distinguishable from a skipped non-candidate. The same load-time inspection used by Action Fusion warns when no configured tool is named `bash`; inner `bash` results from a flat-tool harness still count, because that tool remains in `getAllTools()` even when it is not model-visible.
 
 All persistent paths use `SessionManager.getSessionDir()` and `getSessionId()`, which are present in both the fork and Pi 0.84.2. SoL-Pi creates no configurable storage-path surface.
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -41,9 +41,9 @@ This preflight does not make every valid SoL-Pi configuration all-enabled. Witho
 
 ## Feature behavior
 
-- `actionFusion`: registers SoL-Pi replacements for Pi's `edit` and `write` tools.
+- `actionFusion`: registers SoL-Pi replacements for Pi's `edit` and `write` tools. Fusion is reachable only when a model-visible tool exposes `then_run`. A later-loaded extension that replaces that surface (for example a code-mode harness with a single `fabric_exec` tool) leaves Action Fusion registered but inert; SoL-Pi records a session diagnostic and, in TUI mode, shows a warning.
 - `observationPack`: registers `obs_recall` and a provider-context projection handler.
-- `evidencePreservingReducer`: registers a `tool_result` handler and delegates long diagnostic-log reduction to the configured reducer provider/model.
+- `evidencePreservingReducer`: registers a `tool_result` handler and delegates long diagnostic-log reduction to the configured reducer provider/model. A recovered body under 4 096 bytes is declined with a `source-under-min-bytes` journal entry rather than silently. If no configured tool reports as `bash`, SoL-Pi records the same style of load-time diagnostic used for Action Fusion.
 - `evidencePreservingReducerProvider`: provider namespace used to resolve the reducer model through Pi's model registry.
 - `evidencePreservingReducerModel`: model id used for Evidence-Preserving Reducer.
 - `onlineContextCompact`: registers `update_plan` and boundary-driven native compaction after the other SoL-Pi context transformers.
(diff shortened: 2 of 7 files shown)


gaoanze888 (changes requested):
Thanks for turning #22 into an actionable diagnostic surface. I ran `npm run check` locally (19 files / 149 tests) and the implementation is well isolated, but I found three reachability/lifecycle cases that should be resolved before this becomes authoritative:

1. **Reducer false positive when fused `edit`/`write` is reachable without `bash`.** `reducibleToolResult()` accepts either an outer `bash` result *or* an `edit`/`write` result carrying a `then_run` marker. The diagnostic currently checks only whether `surface.all` contains a tool named `bash`, so a valid active fused `edit` surface with no configured `bash` is reported as reducer-inert. A focused counterexample against this head fails:
   ```ts
   reachabilityFindings(
     config({ actionFusion: true, evidencePreservingReducer: true }),
     surface([{ name: "edit", parameters: thenRunParameters }]),
   ) // returns bash-tool-absent, expected []
   ```
   Please derive reducer reachability from the production candidate contract: reachable `bash`, or an active SoL-Pi fused `edit`/`write` path.

2. **Action Fusion false negative from an unrelated `then_run`.** `actionFusionReachable()` accepts any active tool exposing a `then_run` property. An unrelated extension tool (for example `fabric_exec`) can expose its own field while SoL-Pi's `edit`/`write` are inactive or replaced; that suppresses the warning although Action Fusion cannot execute. Please bind this check to active `edit`/`write`, ideally also using `sourceInfo` to distinguish the SoL-Pi registrations.

3. **Persisted warning and in-memory dedupe diverge after tree navigation.** `appendEntry()` writes beneath the current leaf, while `warned` and `probed` live for the whole extension instance. If the user navigates to an ancestor/other branch, the durable warning can disappear from the active branch, but the closure prevents it from being emitted there. A `session_tree` re-evaluation (with branch-aware dedupe or re-append semantics) would keep the diagnostic available where it is relevant.

There is also an ordering edge: the immediate `session_start` probe can persist a warning before later extensions finish their own `session_start` handlers, and the first `before_agent_start` probe cannot retract that entry. Consider making the final pre-request surface the authoritative persistence point rather than persisting an early provisional result.

The `source-under-min-bytes` journal itself looks appropriately metadata-only, and the PR correctly avoids claiming to make Action Fusion work through a flat tool surface. The blocker here is diagnostic accuracy, not that non-goal.

KayanoLiam:
@gaoanze888 Thanks for the concrete counterexamples. Addressed this review in **bdcd78de64197e2b681f26383f5426948ad1578f**, pushed to this same PR.

1. **Reducer paths:** the check now accepts configured `bash` OR active SoL-Pi fused `edit`/`write`, matching the candidate result kinds. Added regressions for both fused tools without bash. Configured inactive bash is retained as a possible inner harness path; the diagnostic does not guarantee that a harness forwards its results.
2. **Unrelated `then_run`:** active names and property presence alone no longer qualify. Action Fusion tracks the parameter objects it registers in a WeakMap, and the probe checks that registration identity plus the matching tool name. Pi exposes these objects through `getAllTools()`. This avoids source-path heuristics and rejects unrelated `fabric_exec`, replacement `edit`, and replacement `write` schemas, all covered by tests. This is registration identification, not an integrity guarantee against an extension deliberately reusing the original parameter object while replacing execution.
3. **Tree navigation/dedupe:** removed extension-lifetime `warned`/`probed` state. Each request consults `sessionManager.getBranch()` for its durable warning. Navigation to a branch without the entry causes a re-append on the next request; returning to a branch with the entry does not duplicate it. Added a regression for both directions. Re-evaluation is intentionally at the next request, not immediately during `session_tree`.
4. **Ordering:** no provisional session-start warning is persisted. The authoritative probe runs on every `before_provider_request`, after session-start and agent-start setup. Added a later session-start registration regression and a healthy-first-request/later-unreachable regression.

### Local verification

On the project's pinned **Pi 0.84.2**:
- `npx vitest run tests/reachability.test.ts tests/action-fusion.test.ts tests/evidence-preserving-reducer.test.ts` — PASS, 49 tests.
- `npm run check` — PASS: `tsc --noEmit`, all 157 tests / 19 files, and `npm pack --dry-run`.
- `node scripts/check-pi-compat.mjs` — PASS (public API smoke check, not a version-matrix test).
- `git diff --check` and `git diff --cached --check` — PASS. The repository has no separate formatter/lint command configured; reviewed the final diff for scope and existing tool behavior.

Additional **Pi 0.81.1** verification in an isolated temporary checkout (four Pi packages installed at 0.81.1; no dependency changes in this PR):
- Same targeted Vitest command — PASS, 49 tests.
- `npm run check` — **FAIL**: typecheck passes, 155 tests pass, 2 fail in `tests/online-context-compact-agent-session.test.ts:157` (provider call counts 4 vs expected 3, and 6 vs expected 5). The chained pack step therefore did not execute.
- `node scripts/check-pi-compat.mjs` and a separate `npm pack --dry-run` — PASS.
- Reproduced both exact failures with `npx vitest run tests/online-context-compact-agent-session.test.ts` in isolated copies of **both** previous PR head `17df74c` and upstream base `d7ecfc0`, using the same 0.81.1 dependencies. This is an existing compatibility issue, not a flaky-test claim; it remains unresolved/outside this diagnostic fix. The old-version dependency installation also reported 4 moderate and 2 high audit findings; dependencies were not changed here.

An initial typecheck caught readonly-array assignment in my new test fixture; corrected it and reran the checks above.

### GitHub CI

No checks are reported for this PR; there is no PR test workflow in the repository. Local results above are **not** GitHub CI results. In particular, I am not claiming the full 0.81.1 suite passes. Please re-review the diagnostic changes.

gaoanze888 (commented):
Re-reviewed exact head `bdcd78de64197e2b681f26383f5426948ad1578f`. The previous blockers are resolved:

- Action Fusion reachability is tied to the registered SoL-Pi tool parameter identity, so unrelated or replacement tools with the same name cannot forge `then_run` reachability.
- Reducer reachability matches the real candidate contract (`bash` or active fused `edit`/`write`), and diagnostics are emitted from current branch state rather than provisionally at session start.
- The `minBytes` fallback records a metadata-only `source-under-min-bytes` journal entry with actual size/threshold and does not invoke the reducer model.

`npm run check` passed typecheck, all 157 tests, and package dry-run; `git diff --check` is clean. I found no remaining correctness blocker in this PR's scope.

gaoanze888 (approved):
Re-reviewed exact head bdcd78de64197e2b681f26383f5426948ad1578f. The prior requested changes are resolved, and npm run check passes typecheck, all 157 tests, and package dry-run. Approving.

Owen718:

PR #24, opened 2026-09-12
fix(observation-pack): add best-effort Windows symlink fallback without O_NOFOLLOW (#16)

### LINKED ISSUE #16: Windows support: O_NOFOLLOW is undefined, so the ObservationPack symlink guard silently no-ops; 4 tests fail on Windows
## Summary

On Windows, `npm test` fails 4 of 134 tests. Three are POSIX-only assumptions in the test suite, but one reflects a runtime difference that **silently disables a security guard**:

`fs.constants.O_NOFOLLOW` is `undefined` on Windows, and `O_RDONLY | undefined === 0`, so `READ_OBJECT_FLAGS` in the ObservationPack provider degrades to a plain `O_RDONLY` open. The "refusing symlinks" guarantee never takes effect on Windows, and nothing throws at load time to signal it.

## Environment

| Component | Version |
|---|---|
| OS | Windows 10.0.19045 (22H2), x64 |
| Node.js | v22.22.2 |
| npm | 10.9.7 |
| Pi | 0.84.2 (the exact tested release) |
| SoL-Pi | `main` @ `22277b7e0c3c46ba1259a6687f31fe39ade421a5` |

## Reproduction

```bash
git clone https://github.com/NVlabs/SoL-Pi
cd SoL-Pi
npm ci --ignore-scripts
npm test
```

Result: **Test Files 3 failed | 14 passed (17); Tests 4 failed | 130 passed (134)**

For context, everything else passes on the same machine: `npm run typecheck`, `npm pack --dry-run`, `node scripts/check-pi-compat.mjs`, and `npx vitest run tests/all-mechanisms.test.ts` (4/4). Pi starts with both enabled mechanisms and reports no extension error. So this is limited to the test suite plus one Windows runtime guard — the packaging itself is fine.

## Failure 1–2: `tests/package.test.ts` — `spawnSync("npm", ...)` is `ENOENT` on Windows

```js
// tests/package.test.ts:12-19
const result = spawnSync("npm", ["pack", "--dry-run", "--json"], {
	cwd: process.cwd(),
	encoding: "utf8",
});
if (result.status !== 0) throw new Error(result.stderr || result.stdout);
```

On Windows the `npm` on PATH is `npm.cmd`. `child_process.spawnSync` without `shell: true` does not apply PATHEXT resolution, so the spawn fails with `ENOENT`. `result.status` is then `null`, `result.stdout`/`result.stderr` are empty strings, and the thrown `Error` ends up with an empty message:

```
{
  message: '',
  stacks: [ { method: 'packedFiles', file: '.../tests/package.test.ts', line: 19, column: 33 } ]
}
 ❯ packedFiles tests/package.test.ts:19:33
 ❯ tests/package.test.ts:33:17   // and :41:17
```

The underlying command demonstrably works on the same machine — only the spawn call is wrong:

```console
$ npm pack --dry-run --json > /tmp/pack.out; echo $?
0
$ head -c 40 /tmp/pack.out
[
  {
    "id": "sol-pi@0.1.0",
```

Verified root cause:

```console
$ node -e "const r=require('child_process').spawnSync('npm',['--version']);console.log(r.status, r.error && r.error.code)"
null ENOENT
```

**Suggested fix:** call `npm.cmd` on Windows, resolve the executable via `process.env.npm_execpath`, or pass `shell: true`.

## Failure 3: `tests/evidence-preserving-reducer.test.ts:296` — POSIX file mode assertion

```js
expect((await stat(sourcePath)).mode & 0o777).toBe(0o600);
```

`archive.ts:38` writes the archive with `{ mode: 0o600 }`. NTFS does not surface POSIX permission bits through `stat`, so the mode has no effect and the assertion receives `438` (0o666) instead of `384` (0o600).

**Suggested fix:** skip the mode assertion when `process.platform === "win32"`.

## Failure 4: `tests/observation-pack.test.ts:258` — the symlink guard does not fail closed

The test replaces the archived object path with a symlink to another file and expects `obs_recall` to reject with `ELOOP`:

```js
await writeFile(target, "target bytes must not be recalled");
await rm(path);
await symlink(target, path);

await expect(
	pi.tool("obs_recall").execute("recall-1", { id, offset: 0 }, undefined, undefined, fakeContext(sessionDir)),
).rejects.toMatchObject({ code: "ELOOP" });
```

On Windows, recall **succeeds** and returns the symlink target's contents instead:

```json
{ "details": { "bytes": 33, "eof": true, "lines": 1, "offset": 0, "nextOffset": 33 },
  "content": [ { "type": "text", "text": "[obs_recall ...] target bytes must not be recalled" } ] }
```

### Root cause — `src/sol-pi/extensions/observation-pack/observation.ts:21-22`

```js
const READ_OBJECT_FLAGS = constants.O_RDONLY | constants.O_NOFOLLOW;
const CREATE_OBJECT_FLAGS = constants.O_WRONLY | constants.O_CREAT | constants.O_EXCL | constants.O_NOFOLLOW;
```

```console
$ node -e "const c=require('fs').constants;console.log('O_NOFOLLOW =',c.O_NOFOLLOW);console.log('read flags =',c.O_RDONLY | c.O_NOFOLLOW);console.log('create flags =',c.O_WRONLY|c.O_CREAT|c.O_EXCL|c.O_NOFOLLOW)"
O_NOFOLLOW = undefined
read flags = 0            # exactly O_RDONLY — the guard is gone
create flags = 1281       # O_WRONLY|O_CREAT|O_EXCL only
```

`O_NOFOLLOW` is unavailable in Node on Windows, and `undefined` contributes `0` to a bitwise OR. So the flag silently vanishes from both constants rather than raising an error — the failure is invisible until the test above exercises it.

Worth noting the module already handles this correctly one layer up: `observation.ts:126-129` validates the archive *directory* with an explicit `lstat` + `isSymbolicLink()` check, which is cross-platform. The file-open path is the remaining gap.

### Impact

The guarantee documented at `observation.ts:119-121` ("refusing symlinks") does not hold on Windows. Eligibility still requires a user-replaced archive object path, so this is not remotely triggerable, but the local defense that fails closed on POSIX fails open on Windows.

### Suggested fix

Use the same approach as the directory check, which already works cross-platform:

- detect `constants.O_NOFOLLOW === undefined` once and either refuse to start with a clear error, or
- fall back to an `lstat`/`fstat` verification after opening (reject `isSymbolicLink()`, require a regular file), mirroring the existing directory check.

## Requested scope

1. Make the ObservationPack symlink guard either cross-platform or explicitly unsupported on Windows (highest priority — currently a silent no-op).
2. Make the `npm` invocation in `tests/package.test.ts` cross-platform.
3. Skip or document the POSIX-only assertions (`0o600` mode, `ELOOP` on symlinks) on Windows.

Happy to open a PR for any of these. Tested against the pinned Pi 0.84.2 with no changes to upstream Pi, and the SoL-Pi checkout was left unmodified.

Rayan-and-beyond:
Fixes #16.

## What was broken

`constants.O_NOFOLLOW` is `undefined` on Windows, and `undefined`
contributes `0` to a bitwise OR — so the ObservationPack symlink refusal
silently never took effect there (recall succeeded on symlinked objects
instead of ELOOP). Two POSIX-only test assumptions (`spawnSync("npm")`,
a `0o600` mode assertion) failed 3 more tests on Windows.

## What changed

- New exported `assertNotSymlink()`: `lstat` pre-check giving the ELOOP
  the open itself would have raised. Consulted before all three guarded
  opens only where `O_NOFOLLOW` is unavailable; POSIX keeps the atomic
  kernel path untouched. Missing paths fall through so `open()` still
  reports ENOENT; other `lstat` errors rethrow.
- `package.test.ts`: Windows runs the static command line through
  `cmd.exe /d /s /c` (direct `.cmd` spawn can EINVAL under current Node
  hardening; all arguments are literals).
- `evidence-preserving-reducer.test.ts`: the `0o600` assertion is skipped
  on win32 only (NTFS has no POSIX bits); every other assertion still runs.
- New `observation-symlink-guard.test.ts`: symlink→ELOOP, regular→ok,
  absent→ok, EACCES rethrown, plus a POSIX-only pin that the kernel path
  refuses atomically without the fallback. The existing ELOOP recall test
  is unchanged.

## Validation

- New suite 5/5; related suites green; existing ELOOP recall test untouched.
- `npm run check` green (typecheck + 19 files / 144 tests + pack).
- `npm audit --audit-level=high`: exit 0. `check-pi-compat.mjs`: exit 0.
- Windows execution still needs confirmation (no Windows CI here): the
  fallback helper is fully covered and POSIX is byte-identical, but the
  three gated call sites want a Windows run — the existing ELOOP recall
  test is the one to watch.\n\n## Scope note (best-effort, not atomic)\n\nOn Windows this is a best-effort pre-check, not the atomic guarantee POSIX `O_NOFOLLOW` provides: `lstat` followed by `open` leaves a TOCTOU window (a final component can be swapped before open; a checked directory can become a junction before create). It rejects an already-present stable symlink, consistent with SECURITY.md (SoL-Pi is not a sandbox or permission boundary). POSIX path unchanged and atomic.

### DIFF
--- a/src/sol-pi/extensions/observation-pack/observation.ts
+++ b/src/sol-pi/extensions/observation-pack/observation.ts
@@ -20,6 +20,33 @@ const CHARS_PER_TOKEN = 4;
 const OBSERVATION_ID_PATTERN = /^obs_[a-f0-9]{24}$/u;
 const READ_OBJECT_FLAGS = constants.O_RDONLY | constants.O_NOFOLLOW;
 const CREATE_OBJECT_FLAGS = constants.O_WRONLY | constants.O_CREAT | constants.O_EXCL | constants.O_NOFOLLOW;
+const O_NOFOLLOW_SUPPORTED = typeof constants.O_NOFOLLOW === "number";
+
+/**
+ * Best-available symlink pre-check for platforms without O_NOFOLLOW.
+ * Node leaves O_NOFOLLOW `undefined` on Windows, and `undefined`
+ * contributes 0 to the flag words above — so without this the symlink
+ * refusal would silently vanish there (#16). Gives the ELOOP the open
+ * itself would have raised. This is a pre-open lstat, not the atomic
+ * guarantee O_NOFOLLOW provides on POSIX (no pure-Node primitive on
+ * Windows preserves it); on platforms with O_NOFOLLOW the open keeps
+ * refusing atomically and this is never consulted.
+ */
+export async function assertNotSymlink(path: string): Promise<void> {
+	let stats;
+	try {
+		stats = await lstat(path);
+	} catch (error) {
+		// Absent: let open() report ENOENT itself. Anything else (EACCES,
+		// for example) is real and must surface, not vanish.
+		if (error instanceof Error && "code" in error && error.code === "ENOENT") return;
+		throw error;
+	}
+	if (!stats.isSymbolicLink()) return;
+	throw Object.assign(new Error(`Stored observation is not a regular file: ${path}`), {
+		code: "ELOOP",
+	});
+}
 
 /**
  * Receipts from the evidence-preserving reducer are already a reduction of a
@@ -130,10 +157,12 @@ export async function ensureStored(observation: Observation): Promise<void> {
 
 	let handle: FileHandle | undefined;
 	try {
+		if (!O_NOFOLLOW_SUPPORTED) await assertNotSymlink(observation.filePath);
 		handle = await open(observation.filePath, CREATE_OBJECT_FLAGS, 0o600);
 		await handle.writeFile(observation.text, { encoding: "utf8" });
 	} catch (error) {
 		if (!(error instanceof Error) || !("code" in error) || error.code !== "EEXIST") throw error;
+		if (!O_NOFOLLOW_SUPPORTED) await assertNotSymlink(observation.filePath);
 		const existingHandle = await open(observation.filePath, READ_OBJECT_FLAGS);
 		try {
 			const existing = await existingHandle.stat();
@@ -215,6 +244,7 @@ export async function readRecallChunk(
 	offset: number,
 	limits: { readonly maxBytes: number; readonly maxLines: number },
 ): Promise<RecallChunk> {
+	if (!O_NOFOLLOW_SUPPORTED) await assertNotSymlink(path);
 	const handle = await open(path, READ_OBJECT_FLAGS);
 	try {
 		const fileStats = await handle.stat();
--- a/tests/evidence-preserving-reducer.test.ts
+++ b/tests/evidence-preserving-reducer.test.ts
@@ -293,7 +293,11 @@ describe("evidence-preserving reducer", () => {
 		const localSourcePath = relative(join(runtimeRoot(context), "evidence-preserving-reducer"), sourcePath);
 		expect(localSourcePath.length > 0 && !localSourcePath.startsWith("..") && !isAbsolute(localSourcePath)).toBe(true);
 		expect(await readFile(sourcePath, "utf8")).toBe(body);
-		expect((await stat(sourcePath)).mode & 0o777).toBe(0o600);
+		// NTFS does not surface POSIX permission bits, so the 0o600 mode has
+		// no effect there (#16). Every other assertion in this test still runs.
+		if (process.platform !== "win32") {
+			expect((await stat(sourcePath)).mode & 0o777).toBe(0o600);
+		}
 		expect(events.filter((entry) => entry.kind === "applied")).toHaveLength(1);
 		expect(notify).toHaveBeenCalledTimes(1);
 		expect(notify.mock.calls[0]?.[0]).toMatch(
--- a/tests/observation-symlink-guard.test.ts
+++ b/tests/observation-symlink-guard.test.ts
@@ -1,0 +1,83 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: MIT
+ */
+import { constants } from "node:fs";
+import { mkdtemp, open, rm, symlink, writeFile } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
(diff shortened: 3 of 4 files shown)


gaoanze888 (changes requested):
Thanks for tackling the Windows gap and for explicitly documenting that `lstat` is not equivalent to POSIX `O_NOFOLLOW`. I found no Blocker/High issue, but two Medium items remain before merge:

1. **The Windows production branches are not yet validated on Windows.** The PR body explicitly says Windows execution still needs confirmation, and this repository has no CI result for the head. On POSIX, `O_NOFOLLOW_SUPPORTED` is true, so the new pre-checks at the three production call sites are skipped. The helper unit tests are useful, but they do not prove the existing end-to-end recall test now rejects a Windows symlink/junction or that `cmd.exe /d /s /c npm pack ...` behaves correctly there. Please provide a native Windows run of at least the existing recall symlink test, the new helper suite, package tests, and preferably `npm run check`.

2. **Please narrow the fail-closed claim to the guarantee this fallback can provide.** `lstat(path)` followed by `open(path, ...)` remains a TOCTOU window on Windows: the final component can be replaced before recall/existing-object open, and a checked directory component can be replaced by a junction before create. Given `SECURITY.md` correctly says SoL-Pi is not a sandbox or permission boundary, a best-effort pre-check is a reasonable compatibility improvement—but the PR title/summary should not imply the atomic fail-closed guarantee provided by POSIX `O_NOFOLLOW`. Please describe it as rejecting an already-present stable symlink (best effort), or use a Windows-native handle/reparse-point strategy if an atomic security guarantee is intended.

The POSIX path remains unchanged and atomic, the ENOENT/error behavior is sensible, and skipping POSIX mode assertions on NTFS is appropriate.

Rayan-and-beyond:
Thanks for the careful review — both points are fair. Wording narrowed (title + scope note: best-effort pre-check, TOCTOU acknowledged, POSIX atomic path untouched). For the Windows run: lining up a native Windows machine now and will report back with the recall test, helper suite, package tests, and full check results.

Rayan-and-beyond:
Thanks for the review. I’m no longer able to satisfy the Windows validation requirement, so I’m going to release this and close the PR.

Appreciate the detailed feedback.

PR #26, opened 2026-09-13
fix(reducer): count a trailing newline as a line terminator, not an extra line

kaluli123123:
Fixes #36

## Summary

`archiveBody()` reports one line too many for every log that ends with a newline, which is essentially every command output.

`lines` is derived from `body.split("\n").length`. The split produces a trailing empty string for a newline-terminated body, so that empty string is counted as a line.

## Reproduction

`main` @ `d7ecfc0`, Node.js `v22.22.2`, pinned Pi `0.84.2`.

```js
import { archiveBody } from "./src/sol-pi/extensions/evidence-preserving-reducer/archive.ts";
const root = await mkdtemp(join(tmpdir(), "probe-"));
for (const body of ["line1\nline2\n", "line1\nline2", "single line\n", "a\n\n"]) {
  console.log(JSON.stringify(body), (await archiveBody(root, body)).lines);
}
```

```text
"line1\nline2\n"   3     <- 2 lines
"line1\nline2"     2
"single line\n"    2     <- 1 line
"a\n\n"            3     <- 2 lines
```

## Why it matters

`ArchiveObject.lines` is not an internal bookkeeping value. It reaches three places a reader can see:

- `reducerInput()` sends it to the reducer model as `source_lines=` alongside `source_bytes=` and `source_sha256=`;
- `receiptText()` writes it into the receipt the frontier agent reads back (`source_lines=`), next to the `readback=use bash with an explicit byte or line range on source_artifact` instruction;
- the `candidate` / journal entries record it as `sourceLines`.

So a 401-line pytest log is described to both models as 402 lines, and a line-range readback derived from that count addresses the wrong tail.

ObservationPack already counts this correctly in `observation.ts:countLines()` — a trailing newline terminates the last line rather than opening a new one. This aligns the reducer with that.

## Change

`archive.ts` counts newline bytes with a trailing-newline-aware helper instead of splitting. One behavioural change, no new dependency, no API change.

## Tests

Two added to `tests/evidence-preserving-reducer.test.ts`:

- a table over `archiveBody()` covering empty, unterminated, terminated, and blank-final-line bodies;
- an end-to-end assertion that a 401-line newline-terminated log produces `source_lines=401` in the emitted receipt.

Red/green verified — reverting `archive.ts` to `split("\n").length` fails both:

```text
AssertionError: expected 2 to be 1
AssertionError: expected 'sol_pi_evidence_receipt_v1\nstatus=fa…' to contain 'source_lines=401'
[...]

PR #27, opened 2026-09-13
fix(reducer): enforce the cargo subcommand restriction in DIAGNOSTIC_COMMAND

### LINKED ISSUE #37: DIAGNOSTIC_COMMAND does not actually restrict cargo to build/test/check
## Summary

The `cargo` alternative in `DIAGNOSTIC_COMMAND` lists `build|test|check`, but the restriction has no effect: every `cargo <anything>` passes the eligibility gate.

```ts
cargo(?:\s+(?:build|test|check))?
```

The subcommand group is optional, so bare `cargo` already satisfies the alternative, and the trailing `(?:\s|$)` of the outer group matches the space that precedes whatever subcommand follows. The group can never change the match outcome.

https://github.com/NVlabs/SoL-Pi/blob/d7ecfc089944f0d04b80122a0a9a6ca0d786f3d0/src/sol-pi/extensions/evidence-preserving-reducer/config.ts#L25-L26

## Reproduction

```js
import { DIAGNOSTIC_COMMAND } from "./src/sol-pi/extensions/evidence-preserving-reducer/config.ts";
for (const c of ["cargo build", "cargo fmt", "cargo publish", "cargo login", "cargo install foo", "cargo"])
  console.log(c.padEnd(20), DIAGNOSTIC_COMMAND.test(c));
```

## Actual behavior

```text
cargo build          true
cargo fmt            true     <- not build|test|check
cargo publish        true     <- not build|test|check
cargo login          true     <- not build|test|check
cargo install foo    true     <- not build|test|check
cargo                true     <- no subcommand at all
```

For comparison, the neighbouring `python(?:3)?\s+-m\s+(?:pytest|unittest|py_compile)` alternative has no optional group and does restrict correctly.

## Impact

`DIAGNOSTIC_COMMAND` is the eligibility gate in `reduceToolResult()`. Passing it is what makes a tool result's body eligible to be archived and sent to the configured reducer model.

`SECURITY.md` frames remote reduction as something to enable deliberately for diagnostic logs. `cargo publish` / `cargo login` / `cargo install` output is registry and credential-adjacent traffic rather than a build or test log, and `LIKELY_SECRET` is documented as "a precaution rather than a complete secret scanner".

There is a cost side too: every false positive spends a reducer model call on output the mechanism was never meant to reduce.

## Expected behavior

The cargo alternative should admit only its diagnostic subcommands, as the listed alternatives suggest was intended. A toolchain selector (`cargo +nightly test`) should stay eligible, since it matches today.

I have a focused fix with regression tests ready.

kaluli123123:
Fixes #37

## Summary

The `cargo` alternative in `DIAGNOSTIC_COMMAND` lists `build|test|check`, but the restriction has no effect: every `cargo <anything>` passes the gate.

```ts
cargo(?:\s+(?:build|test|check))?
```

The subcommand group is optional, so bare `cargo` already satisfies the alternative, and the trailing `(?:\s|$)` of the outer group matches the space that precedes whatever subcommand follows. The group can never change the match outcome — it is dead syntax that reads like a restriction.

## Reproduction

`main` @ `d7ecfc0`, Node.js `v22.22.2`.

```js
import { DIAGNOSTIC_COMMAND } from "./src/sol-pi/extensions/evidence-preserving-reducer/config.ts";
for (const c of ["cargo build", "cargo fmt", "cargo publish", "cargo login", "cargo install foo", "cargo"])
  console.log(c.padEnd(20), DIAGNOSTIC_COMMAND.test(c));
```

```text
cargo build          true
cargo fmt            true     <- not build|test|check
cargo publish        true     <- not build|test|check
cargo login          true     <- not build|test|check
cargo install foo    true     <- not build|test|check
cargo                true     <- no subcommand at all
```

For comparison, the neighbouring `python(?:3)?\s+-m\s+(?:pytest|unittest|py_compile)` alternative has no optional group and does restrict correctly.

## Why it matters

`DIAGNOSTIC_COMMAND` is the eligibility gate in `reduceToolResult()`. Passing it is what makes a tool result's body eligible to be archived and sent to the configured reducer model. `SECURITY.md` frames remote reduction as something to enable deliberately for diagnostic logs; `cargo publish` / `cargo login` / `cargo install` output is registry and credential-adjacent traffic, not a build or test log, and `LIKELY_SECRET` is documented as "a precaution rather than a complete secret scanner".

The cost side matters too: every false positive spends a reducer model call on output the mechanism was never meant to reduce.

## Change

```diff
-cargo(?:\s+(?:build|test|check))?
+cargo(?:\s+\+[\w.-]+)?\s+(?:build|test|check)
```

The subcommand is now required. The optional `+toolchain` selector keeps `cargo +nightly test` eligible — without it, requiring the subcommand would newly exclude toolchain-pinned invocations that match today.

No other alternative in the pattern is touched.

## Tests

Two added to `tests/evidence-preserving-reducer.test.ts`:

- a table asserting `cargo build` / `cargo test` / `cargo check` / `cargo build --release` / `cargo +nightly test` / `cd repo && cargo build` still match, and `cargo` / `cargo fmt` / `cargo clippy` / `cargo run` / `cargo publish` / `cargo login` / `cargo install cargo-nextest` do not;
- an end-to-end assertion that a `cargo publish` result returns `undefined` from the `tool_result` handler, makes no reducer model call, and writes no journal entry.

Red/green verified — with `config.ts` reverted:

```text
AssertionError: expected [ 'cargo', true ] to deeply equal [ 'cargo', false ]
AssertionError: expected "vi.fn()" to not be called at all, but actually been called 1 times
Tests  2 failed | 15 passed (17)
```

```text
```

The existing `keeps the diagnostic command trigger generic` test (`cargo test --all` → true) still passes unchanged.


kaluli123123:
Merge note: #26, #27, and #28 each insert new `it(...)` blocks at the same anchor in `tests/evidence-preserving-reducer.test.ts`, so whichever lands second needs a one-line rebase there. No source files overlap between them — `archive.ts` vs. the two `config.ts` regexes are independent. I will rebase mine on request, in whatever order suits you.

gaoanze888 (changes requested):
The core restriction fix is valid, and the current tests/full suite pass locally (18 files / 141 tests). One functional regression remains: the new pattern requires the subcommand immediately after an optional `+toolchain`, so valid diagnostic invocations with Cargo global options are rejected. For example, this head returns false for both:

```text
cargo --locked test
cargo --color always check
```

Both are accepted Cargo command lines and still produce the diagnostic output this reducer is meant to handle. Please allow a controlled set of pre-subcommand global options (including options with values), or use a small argv-level parser, while continuing to reject `publish`, `login`, `install`, etc. Add positive regressions for at least the two cases above and keep the end-to-end negative test.

gaoanze888 (changes requested):
Re-reviewed exact head `da9379192df52bd296d051025f6c0ac638f5ed7c`. The originally reported `--locked` / `--color always` cases now match and the full local suite passes (18 files / 142 tests), but the expanded regex introduces two serious issues:

1. The nested repeated option pattern has catastrophic backtracking on a non-match. Repeated `--color` tokens took roughly 60 ms at 30 tokens, 658 ms at 35, and 7.3 seconds at 40 on the review environment, blocking the Node event loop. Please replace this with bounded token parsing or an unambiguous expression.
2. The pattern can skip a non-diagnostic Cargo subcommand and match a later `cargo test` substring in its arguments. `cargo login cargo test`, `cargo install cargo test`, and `cargo publish --package cargo test` all return true, violating the diagnostic-only gate.

There are also grammar inconsistencies: valid `--color=always` / `--config=...` forms are rejected, while incomplete `cargo --color test` / `cargo --config test` can be accepted by backtracking. Please use argv-level parsing for the Cargo arm, anchor it to the actual invocation/subcommand, and add these positive/negative plus bounded-time regressions.

kaluli123123:
Addressed in `677f86f`.

The Cargo gate now uses bounded shell-token parsing instead of the repeated regex. It identifies the actual Cargo subcommand, accepts controlled global options in both separated and `--option=value` forms, and rejects non-diagnostic subcommands even when their arguments contain a later `cargo test` substring. The regression coverage includes the reported valid, invalid, incomplete-option, and bounded-time cases.

Local validation passes: 143 tests, the focused reducer suite, all-mechanisms checks, Pi compatibility, `git diff --check`, and the high-severity audit gate. Please re-review the current head.

gaoanze888 (changes requested):
Re-reviewed exact head `677f86f182c1e42988fd3386da110590c7c618f5`. The token scanner fixes the prior catastrophic backtracking and ordinary suffix false positives, but quoted empty arguments are dropped, so option values can still shift into the subcommand position. Reproduction: `DIAGNOSTIC_COMMAND.test("cargo -C [0m[0m publish test")` is not useful here; the literal shell case `cargo -C '\' publish test` returns true even though Cargo parses the empty string as `-C`’s value and then runs/rejects `publish`, not diagnostic `test`. The same issue affects `cargo --config '\' login test` and `cargo -Z '\' login test`. Please track whether a token has started independently of `token.length`, preserve quoted empty argv entries, and add these regressions. Also consider stopping unquoted `#` comments so `cargo login # ; cargo test` is not treated as an executable second command. Current-main integration otherwise passes.

gaoanze888:
Correction to the quoted examples in my review body (shell escaping was mangled while posting): the exact reproductions are:

```text
cargo -C '' publish test
cargo --config '' login test
cargo -Z '' login test
```

All three currently return `true`. The empty quoted string is a real argv entry and must be consumed as the option value; dropping it shifts the later `test` token into the apparent subcommand position.

kaluli123123:
Addressed in `46042dc`.

The shell scanner now tracks token presence independently from token text, so quoted empty arguments remain real argv entries and are consumed as Cargo option values. It also stops at an unquoted shell comment that begins at a token boundary, preventing commented-out separators from exposing a later diagnostic command. The exact `-C`, `--config`, `-Z`, and comment reproductions are covered.

Validation passes: `npm run check` (18 files / 144 tests), Pi compatibility, all-mechanisms (4/4), `git diff --check`, and the high-severity audit gate. The audit still reports the existing two moderate Vitest advisories. Please re-review the current head.

gaoanze888 (changes requested):
The quoted-empty argv bug is fixed on exact head `46042dc98d2b`: the three `-C` / `--config` / `-Z` reproductions now return false, and the focused suite passes 20/20. Two parser gaps still make the overall diagnostic gate accept commands that do not execute a diagnostic:\n\n1. `NON_CARGO_DIAGNOSTIC_COMMAND` runs on the raw command before tokenization, so `cargo login # pytest -q` and `echo '# pytest -q'` both return true even though `pytest` is comment/quoted data.\n2. Inside double quotes the scanner removes a backslash before every character. POSIX only treats backslash specially there before `$`, backtick, `"`, `\\`, or newline. `cargo "\\test"` is therefore parsed as subcommand `test` and returns true, while the shell argv subcommand is `\\test`.\n\nI reproduced all three directly against exported `DIAGNOSTIC_COMMAND`. Please route non-Cargo detection through the same parsed segments/tokens and preserve backslashes inside double quotes except for shell-defined escapable characters, with these negative regressions.

Owen718:

gaoanze888 (approved):
Re-reviewed at exact head b73c99a425.

Both parser gaps I flagged are closed. The `#` comment handling now only starts a comment at a token boundary, so `cargo login # pytest -q` and `echo '# pytest -q'` both return false (comment/quoted data is no longer treated as an executable second command). Inside double quotes, the backslash is now preserved except before the shell-escapable set (`$`, backtick, `"`, `\\`), so `cargo "\\test"` no longer mis-parses the subcommand as `test`.

Verified directly against the exported `DIAGNOSTIC_COMMAND` (all three former false-positives now false; `cargo build`/`cargo test`/`pytest` still true; `cargo publish`/`cargo login` still false) and the focused reducer suite passes 33/33. Approving.

PR #28, opened 2026-09-13
fix(reducer): detect password and PEM private-key material before remote reduction

### LINKED ISSUE #38: Likely-secret detector misses password assignments and PEM private keys
## Summary

`LIKELY_SECRET` is the last gate before a log is sent to the configured reducer model. It misses two shapes that appear in ordinary build, test, and deploy output: the password family, and PEM private-key blocks.

```ts
/(?:api[_-]?key|authorization|bearer|access[_-]?token|secret)[^\n]{0,32}[=:][^\n]+/i
```

https://github.com/NVlabs/SoL-Pi/blob/d7ecfc089944f0d04b80122a0a9a6ca0d786f3d0/src/sol-pi/extensions/evidence-preserving-reducer/config.ts#L29

## Reproduction

```js
import { LIKELY_SECRET } from "./src/sol-pi/extensions/evidence-preserving-reducer/config.ts";
for (const b of [
  "api_key=abc123",
  "PGPASSWORD=hunter2",
  '  "password": "hunter2"',
  "-----BEGIN OPENSSH PRIVATE KEY-----",
]) console.log(String(LIKELY_SECRET.test(b)).padEnd(6), b);
```

## Actual behavior

```text
true   api_key=abc123
false  PGPASSWORD=hunter2                    <- eligible for the remote call
false    "password": "hunter2"               <- eligible for the remote call
false  -----BEGIN OPENSSH PRIVATE KEY-----   <- eligible for the remote call
```

## Impact

A concrete path: a failing `pytest` run over deploy or signing code that prints the key material it loaded. `pytest` is in `DIAGNOSTIC_COMMAND`, the body clears `minBytes`, so the only thing between that log and `callReducer()` is this regex.

`SECURITY.md` already says the detector "is a precaution rather than a complete secret scanner", so this is not a contract violation — but these two shapes are common enough to be worth catching.

## Expected behavior

Password/passphrase assignments and PEM private-key headers should fall back instead of being sent.

Worth noting the trade-off in the other direction: reusing the existing `{0,32}` slack for the password family would make `FAILED tests/test_auth.py::test_password_reset` match, and a false positive costs a real reduction on exactly the logs this mechanism exists for. A tighter window before the assignment avoids that:

```text
SKIP-REDUCTION "PGPASSWORD=hunter2"
SKIP-REDUCTION "db_password: hunter2"
SKIP-REDUCTION "  \"password\": \"hunter2\""
SKIP-REDUCTION "-----BEGIN OPENSSH PRIVATE KEY-----"
reduce-ok      "-----BEGIN CERTIFICATE-----"
reduce-ok      "FAILED tests/test_auth.py::test_password_reset"
reduce-ok      "assert password_hash == expected_hash"
reduce-ok      "  password_policy_enabled = True"
```

I have a focused fix with regression tests covering both directions ready.

kaluli123123:
Fixes #38

## Summary

`LIKELY_SECRET` is the last gate before a log is sent to the configured reducer model. It misses two shapes that appear in ordinary build, test, and deploy output: the password family, and PEM private-key blocks.

```ts
/(?:api[_-]?key|authorization|bearer|access[_-]?token|secret)[^\n]{0,32}[=:][^\n]+/i
```

## Reproduction

`main` @ `d7ecfc0`, Node.js `v22.22.2`.

```js
import { LIKELY_SECRET } from "./src/sol-pi/extensions/evidence-preserving-reducer/config.ts";
for (const b of [
  "api_key=abc123",
  "PGPASSWORD=hunter2",
  '  "password": "hunter2"',
  "-----BEGIN OPENSSH PRIVATE KEY-----",
]) console.log(String(LIKELY_SECRET.test(b)).padEnd(6), b);
```

```text
true   api_key=abc123
false  PGPASSWORD=hunter2                    <- eligible for the remote call
false    "password": "hunter2"               <- eligible for the remote call
false  -----BEGIN OPENSSH PRIVATE KEY-----   <- eligible for the remote call
```

A concrete path to this: a failing `pytest` run over deploy or signing code that prints the key material it loaded. `pytest` is in `DIAGNOSTIC_COMMAND`, the body clears `minBytes`, so the only thing standing between that log and `callReducer()` is this regex.

`SECURITY.md` already says the detector "is a precaution rather than a complete secret scanner", so this does not change the contract — it closes two shapes that are common enough to be worth catching.

## Change

Three alternatives instead of one:

1. the existing keyword family, unchanged, with its `{0,32}` slack;
2. `password|passwd|passphrase|private[ _-]?key` with a **tight** `["'\s]{0,4}` window before the assignment;
3. the PEM header `-----BEGIN(?: [A-Z0-9]+)* PRIVATE KEY-----`, which carries no assignment at all.

The tight window on (2) is deliberate. Reusing `{0,32}` there would make `FAILED tests/test_auth.py::test_password_reset` match, and a false positive costs a real reduction on exactly the logs this mechanism exists for. Verified both directions:

```text
SKIP-REDUCTION "PGPASSWORD=hunter2"
SKIP-REDUCTION "db_password: hunter2"
SKIP-REDUCTION "  \"password\": \"hunter2\""
SKIP-REDUCTION "passphrase = hunter2"
SKIP-REDUCTION "-----BEGIN OPENSSH PRIVATE KEY-----"
SKIP-REDUCTION "-----BEGIN ENCRYPTED PRIVATE KEY-----"
reduce-ok      "-----BEGIN CERTIFICATE-----"
reduce-ok      "FAILED tests/test_auth.py::test_password_reset"
reduce-ok      "assert password_hash == expected_hash"
reduce-ok      "  password_policy_enabled = True"
reduce-ok      "E   AssertionError: expected 4 but received 5"
```

`-----BEGIN CERTIFICATE-----` stays reducible: a public certificate is not key material.

## Tests

Two added to `tests/evidence-preserving-reducer.test.ts`:

- a two-sided table over `LIKELY_SECRET` — the eleven secret shapes above match, the five ordinary diagnostic lines do not;
- an end-to-end assertion that a pytest log containing an OpenSSH private key returns `undefined` from the `tool_result` handler, makes no reducer model call, and journals exactly one `{ kind: "fallback", reason: "likely-secret" }` entry.

Red/green verified — with `config.ts` reverted:

```text
AssertionError: expected [ 'PGPASSWORD=hunter2', false ] to deeply equal [ 'PGPASSWORD=hunter2', true ]
AssertionError: expected "vi.fn()" to not be called at all, but actually been called 1 times
Tests  2 failed | 15 passed (17)
```

```text
```


kaluli123123:
Merge note: #26, #27, and #28 each insert new `it(...)` blocks at the same anchor in `tests/evidence-preserving-reducer.test.ts`, so whichever lands second needs a one-line rebase there. No source files overlap between them — `archive.ts` vs. the two `config.ts` regexes are independent. I will rebase mine on request, in whatever order suits you.

gaoanze888 (changes requested):
The new detector catches several valuable cases, and the current tests/full suite pass locally (18 files / 141 tests). Two security/precision gaps remain in the proposed regex:

1. A standard private-key header is still missed:
   ```text
   -----BEGIN PGP PRIVATE KEY BLOCK-----  => false
   ```
   Please include it and add the end-to-end no-provider-call regression.

2. Common diagnostic assertions become false positives:
   ```text
   assert password == expected           => true
   AssertionError: password = expected   => true
   ```
   These are exactly the auth-test logs the reducer should preserve, not assignments containing a secret value. Please distinguish serialization/assignment forms from equality/assertion diagnostics and add negative tests.

Because this gate controls whether potentially sensitive text leaves the local process, I think both directions should be nailed down before merge rather than relying on the disclaimer alone.

gaoanze888 (changes requested):
Re-reviewed exact head `214447805693dd5cdf027900ef3efaa901b2c8a4`. The previous Python assertion examples are fixed, and locally `npm ci --ignore-scripts`, `npm run check` (18 files / 141 tests), `node scripts/check-pi-compat.mjs`, and `git diff --check` pass. Three concrete gaps remain in this pre-remote-call safety gate:

1. Standard PGP private-key armor is still missed: `-----BEGIN PGP PRIVATE KEY BLOCK-----` returns false, so PGP private material can remain eligible for the remote reducer. Please cover it with both detector and end-to-end no-provider-call tests.
2. Common assertion formats still false-positive, including `AssertionError [ERR_ASSERTION]: password = expected` and `AssertionFailedError: password = expected`. Please broaden the assertion-context exclusion and add negatives.
3. Common space-separated CLI secrets such as `tool --password hunter2` are missed. Please recognize bounded password/passphrase flags with a following value while avoiding ordinary prose.

Because this regex is the last local gate before content is sent remotely, the key-material recall gap is blocking.

kaluli123123:
Addressed in `ae58e2b`.

The local safety gate now detects standard PGP private-key armor and bounded space-separated password/passphrase CLI flags, while excluding the reported Node and Chai assertion contexts. End-to-end regressions verify that matched secrets never invoke the remote provider, alongside negative assertion cases.

Local validation passes: 143 tests, the focused reducer suite, all-mechanisms checks, Pi compatibility, `git diff --check`, and the high-severity audit gate. Please re-review the current head.

gaoanze888 (approved):
Re-reviewed exact head `ae58e2b5f64b226d48f6ef9cc9b36043404d3479`. The updated detector now covers PGP private-key armor and whitespace-form `--password`/`--passphrase` values while excluding the reported Node and Java assertion formats. The tests also verify secret hits preserve the original result without calling the reducer provider. Full checks pass locally (18 files / 143 tests, typecheck, package dry-run), as do Pi compatibility, all-mechanisms, focused reducer tests, and diff-check. No blocking issue found.

Owen718:

PR #29, opened 2026-09-13
fix(observation-pack): fail open on ephemeral sessions instead of erroring on every request

### LINKED ISSUE #39: ObservationPack raises an extension error on every request under pi --no-session
## Summary

With `observationPack: true`, running `pi --no-session` reports an extension error on **every** provider request:

```text
SoL-Pi requires a persistent Pi session directory
```

`observation-pack/index.ts` calls `runtimeRoot(ctx)` at the top of the `context` handler, outside the per-message `try/catch` that the module's own comment describes as fail-open.

https://github.com/NVlabs/SoL-Pi/blob/d7ecfc089944f0d04b80122a0a9a6ca0d786f3d0/src/sol-pi/extensions/observation-pack/index.ts#L137-L139

## Reproduction

Observed on `main` commit `d7ecfc089944f0d04b80122a0a9a6ca0d786f3d0`, Node.js `v22.22.2`, with the pinned Pi `0.84.2` and a `sol-pi.json` that enables `observationPack`.

```bash
pi --no-session --offline --approve
```

Any turn that projects context raises the error above. The same path is reachable in a test with a session manager whose `getSessionDir()` is empty.

## Root cause

`pi --no-session` ("Don't save session (ephemeral)", `cli/args.js:267`) routes through `createSessionManager()` in Pi 0.84.2:

```js
if (parsed.noSession || parsed.help || parsed.listModels !== undefined) {
    return SessionManager.inMemory(cwd, ...);
}
```

and `SessionManager.inMemory()` is `new SessionManager(cwd, "", undefined, false, options)` — session directory `undefined`, session id `""`. `runtimeRoot()` rejects both.

Pi's `emitContext()` catches the throw and calls `emitError({ event: "context", ... })`, so nothing crashes, but the user gets one extension error per provider request for the whole session, and the mechanism is inert regardless.

## Contrast with the sibling mechanism

Evidence-Preserving Reducer already handles exactly this:

```ts
let root: string;
try {
    root = runtimeRoot(context);
} catch {
    return undefined;
}
```

## Expected behavior

ObservationPack should leave the context untouched when there is no persistent directory, consistent with its own fail-open comment and with the reducer's handling. Reporting the reason once per extension instance rather than once per request keeps the decline visible without becoming noise.

`obs_recall` should keep failing loudly: a tool call with no archive is a real error.

I have a focused fix with a regression test ready.

kaluli123123:
Fixes #39

## Summary

With `observationPack: true`, running `pi --no-session` reports an extension error on **every** provider request:

```text
SoL-Pi requires a persistent Pi session directory
```

`observation-pack/index.ts` calls `runtimeRoot(ctx)` at the top of the `context` handler, outside the per-message `try/catch` that the module's own comment describes as fail-open:

```ts
pi.on("context", async (event, ctx: ExtensionContext) => {
    const projected = [...event.messages];
    const root = runtimeRoot(ctx);   // <- throws before any per-message guard
```

## Why it fires

`pi --no-session` ("Don't save session (ephemeral)", `cli/args.js:267`) routes through `createSessionManager()` in Pi 0.84.2:

```js
if (parsed.noSession || parsed.help || parsed.listModels !== undefined) {
    return SessionManager.inMemory(cwd, ...);
}
```

and `SessionManager.inMemory()` is `new SessionManager(cwd, "", undefined, false, options)` — session directory `undefined`, session id `""`. `runtimeRoot()` rejects both.

Pi's `emitContext()` catches the throw and calls `emitError({ event: "context", ... })`, so nothing crashes — but the user gets one extension error per provider request, for the whole session, and the mechanism is silently inert regardless.

## Reproduction

`main` @ `d7ecfc0`, Node.js `v22.22.2`, pinned Pi `0.84.2`, `sol-pi.json` with `observationPack: true`.

```bash
pi --no-session --offline --approve
```

Any turn that projects context raises the error above. The same path is reachable in tests with a session manager whose `getSessionDir()` is empty.

## Contrast with the sibling mechanism

Evidence-Preserving Reducer already handles exactly this, in `evidence-preserving-reducer/index.ts`:

```ts
let root: string;
try {
    root = runtimeRoot(context);
} catch {
    return undefined;
}
```

This brings ObservationPack in line.

## Change

The `context` handler resolves the runtime root inside a `try/catch`, returns `undefined` (context untouched) when there is no persistent directory, and reports the reason **once per extension instance** rather than once per request — so the decline is not silent, but it is also not repeated noise. Everything else in the handler is unchanged.

`obs_recall` is deliberately left as-is: a tool call with no archive should still fail loudly.

## Tests

One added to `tests/observation-pack.test.ts`: three consecutive projections over an ephemeral session return the original body unchanged and log exactly one `[observationpack] disabled for this session` line.

Red/green verified — with `index.ts` reverted:

```text
⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯
Error: SoL-Pi requires a persistent Pi session directory
Tests  1 failed | 14 passed (15)
```

```text
npx vitest run            Test Files 18 passed (18) | Tests 140 passed (140)
```


kaluli123123:
Merge note: #29 and #31 both add tests at the same anchor in `tests/observation-pack.test.ts`, so whichever lands second needs a one-line rebase there. Their source changes do not overlap — #29 touches the `context` handler in `index.ts`, #31 touches `placeholderFor()` in `observation.ts`. Happy to rebase in whichever order you prefer.

gaoanze888 (changes requested):
The fail-open behavior is the right direction and the local suite passes, but `reportedUnavailableRuntime` is scoped to the entire extension instance, not to a session. After the first ephemeral session logs the warning, every later ephemeral session handled by the same extension instance is silently suppressed even though the message says “disabled for this session.”

Please scope/reset the warning per active ephemeral session/context, and add a regression with two distinct ephemeral session managers that each receive exactly one warning. This also textually overlaps #13 in the context handler, so please preserve #13's branch-local send-count state when rebasing/merging.

gaoanze888 (approved):
Re-reviewed exact head `3414a30d3b2f488853daf26ff1fc84bf55e6f7f6`. The prior lifecycle issue is resolved: warning dedupe is now keyed by the session-manager object in a `WeakSet`, so each distinct ephemeral session reports once without retaining ended sessions, while repeated requests remain quiet and context stays unchanged. Locally, `npm ci --ignore-scripts`, `npm run check` (18 files / 140 tests), `node scripts/check-pi-compat.mjs`, and `git diff --check` passed. No blocking issue found. Integration with #13 should preserve both branch-local send counts and ephemeral shutdown fail-open behavior.

Owen718:

PR #30, opened 2026-09-13
perf(reducer): apply the diagnostic-command gate before reading the full-output file

### LINKED ISSUE #40: Reducer reads the whole pi-bash full-output file before the diagnostic-command gate
## Summary

Every truncated bash result reads its entire `pi-bash-*.log` into memory, including results the reducer discards one line later.

`reduceToolResult()` runs the gates in this order:

```ts
const reducible = await reducibleToolResult(event);          // recovers the body, reading the full-output file
if (!reducible || !DIAGNOSTIC_COMMAND.test(reducible.command)) return undefined;   // then decides it is not a candidate
```

https://github.com/NVlabs/SoL-Pi/blob/d7ecfc089944f0d04b80122a0a9a6ca0d786f3d0/src/sol-pi/extensions/evidence-preserving-reducer/index.ts#L64-L65

`reducibleToolResult()` calls `exactBodyFromInline()` unconditionally, which `readFile()`s the untruncated log Pi wrote.

## Reproduction

Observed on `main` commit `d7ecfc089944f0d04b80122a0a9a6ca0d786f3d0`, Node.js `v22.22.2`, macOS.

One `git status --porcelain` result whose full-output file is 100 MB, written as a real `pi-bash-<uuid>.log` inside `os.tmpdir()` so it passes `safePiBashTempPath()`:

```js
const event = {
  type: "tool_result", toolName: "bash", toolCallId: "c1",
  input: { command: "git status --porcelain" },
  content: [{ type: "text", text: "truncated preview" }],
  details: { fullOutputPath: path },
  isError: false,
};
const out = await reducibleToolResult(event);
```

## Actual behavior

```text
result           = body of 104857641 chars
elapsed          = 37.3 ms
heap delta       = 99.8 MiB
```

The command is not in `DIAGNOSTIC_COMMAND`, so the caller discards all of it on the next line.

## Impact

Pi truncates inline output at `DEFAULT_MAX_BYTES` (50 KB) and saves the rest to that file, so the file is present for every large bash result, not just build and test commands: `git status` on a big repo, `cat`, `curl`, `find`, `docker logs`, a dev server's stdout.

There is also a robustness edge: a full-output file above Node's maximum string length makes `readFile(path, "utf8")` throw `RangeError`, and the `tool_result` handler would surface that as an extension error for a command the mechanism never intended to touch. The `catch` in `exactBodyFromInline()` covers the read itself, not the string construction limit.

## Expected behavior

The diagnostic-command gate should be applied as soon as the command is known, before the untruncated body is resolved — for the bash branch and for the fused `edit`/`write` branch. Eligibility should not change; the same commands are admitted, they are just no longer paid for before the decision.

I have a focused change with tests that record which paths are actually read.

kaluli123123:
Fixes #40

## Summary

Every truncated bash result reads its entire `pi-bash-*.log` into memory, including results the reducer discards one line later.

`reduceToolResult()` runs the gates in this order:

```ts
const reducible = await reducibleToolResult(event);          // recovers the body, reading the full-output file
if (!reducible || !DIAGNOSTIC_COMMAND.test(reducible.command)) return undefined;   // then decides it is not a candidate
```

`reducibleToolResult()` calls `exactBodyFromInline()` unconditionally, which `readFile()`s the untruncated log Pi wrote. Pi truncates inline output at `DEFAULT_MAX_BYTES` (50 KB) and saves the rest to that file, so the file is present for every large bash result — `git status` on a big repo, `cat`, `curl`, `find`, `docker logs`, a dev server's stdout — not just for build and test commands.

## Measurement

`main` @ `d7ecfc0`, Node.js `v22.22.2`, macOS. One `git status --porcelain` result whose full-output file is 100 MB:

```text
=====  main  =====
result           = body of 104857641 chars
elapsed          = 37.3 ms
heap delta       = 99.8 MiB

=====  this PR  =====
result           = undefined
elapsed          = 0.5 ms
heap delta       = 0.2 MiB
```

The benchmark writes a real `pi-bash-<uuid>.log` into `os.tmpdir()` so it passes `safePiBashTempPath()`, then calls `reducibleToolResult()` once.

Beyond the cost, this is also a robustness edge: a full-output file above Node's maximum string length makes `readFile(path, "utf8")` throw `RangeError`, and the `tool_result` handler would surface that as an extension error for a command the mechanism never intended to touch. The `catch` in `exactBodyFromInline()` only covers the read itself, not the string construction limit.

## Change

`reducibleToolResult()` applies `DIAGNOSTIC_COMMAND` as soon as the command is known — the bash branch right after reading `event.input.command`, the fused branch right after reading `then_run.command` (which also skips the marker scan). `reduceToolResult()` drops the now-duplicate test and keeps a comment pointing at where the gate lives.

Eligibility is unchanged: the same commands are admitted, they are just no longer paid for before the decision.

## Tests

New `tests/evidence-preserving-reducer-candidate.test.ts`, which wraps `readFile` to record the paths actually read:

- `pytest -q` with a full-output file — body recovered from the file, path recorded as read (the existing behaviour, pinned);
- `git status --porcelain` with the same file — `undefined`, path never read;
- a fused `write` whose `then_run` is `node server.js` — `undefined`, path never read.

```text
AssertionError: expected { …(3) } to be undefined
AssertionError: expected { command: 'node server.js', …(2) } to be undefined
Tests  2 failed | 1 passed (3)
```

```text
npx vitest run            Test Files 19 passed (19) | Tests 142 passed (142)
```


gaoanze888 (approved):
Reviewed the gate move across plain bash and fused edit/write candidates. Non-diagnostic commands now avoid full-output reads without changing any journal/fallback path that previously ran, while eligible diagnostics still recover the exact body. `npm run check` passes locally: 19 files / 142 tests, typecheck, and package dry-run. No blocking issue found.

gaoanze888:
Integration note after reviewing #64: #64 now contains this PR’s early diagnostic-command gate in `candidate.ts`/`index.ts`, plus bounded and stability-checked full-output reads and the expanded candidate regressions. The two heads conflict in all three changed paths and should not both be merged. I recommend taking the more complete #64 and closing this PR as superseded; my approval here remains a statement that this exact head was correct standalone, not a recommendation to duplicate the implementation.

Owen718:

PR #33, opened 2026-09-13
perf(observation-pack): 3.8x faster context projection for large observations

### LINKED ISSUE #43: ObservationPack re-walks whole payloads on every context projection
## Summary

ObservationPack's `context` handler runs before **every** provider request, over every large tool result in the context. For each one it walks the whole payload three more times than the answer requires.

| Call | Cost on a 7.5 MiB payload | What it needs |
|---|---|---|
| `containsReducerReceipt` -> `text.split("\n")` | 8.1 ms, one string per line | whether one line equals a 26-char prefix |
| `countLines` -> `for (const character of text)` | 47.5 ms | the number of `\n` |
| `completeLineExcerpt` -> `text.split(/(?<=\n)/)`, twice | 21.8 ms each, one string per line | at most 512 bytes from each end |
| `hash(text)` (sha256) | 5.5 ms | the content address, unavoidable |

`countLines` is the largest single item: `for..of` over a string iterates Unicode code points, which costs about 5x an indexed scan for an answer that only involves `\n`.

## Reproduction

Observed on `main` commit `d7ecfc089944f0d04b80122a0a9a6ca0d786f3d0`, Node.js `v22.22.2`, macOS. Per-operation numbers above are on 160 000 lines of realistic log text.

End to end through the real `context` handler, one 7.5 MiB observation, 20 projections after the first store:

```text
observation bytes = 7.5 MiB
20 projections = 2238 ms  (111.9 ms per provider request)
```

## Impact

This sits on the critical path in front of every model call, in a mechanism whose stated purpose is making long trajectories cheaper. A session with a few multi-megabyte observations pays it on every request for the rest of the session.

## Expected behavior

All three answers are reachable with `indexOf` / `lastIndexOf` / `charCodeAt` scans that slice only what they keep, with no behaviour change. A prototype of that reaches:

```text
20 projections =  594 ms  (29.7 ms per provider request)
```

3.8x, with identical output on 60 000 randomly assembled payloads across 12 excerpt variants (empty strings, bare newlines, CRLF, astral characters, the receipt prefix in every position).

I have that change plus characterization tests ready.

kaluli123123:
Fixes #43

## Summary

ObservationPack's `context` handler runs before **every** provider request, over every large tool result in the context. For each one it walked the whole payload three more times than the answer required:

| Call | Cost on a 7.5 MiB payload | What it needed |
|---|---|---|
| `containsReducerReceipt` → `text.split("\n")` | 8.1 ms, one string per line | whether one line equals a 26-char prefix |
| `countLines` → `for (const character of text)` | 47.5 ms | the number of `\n` |
| `completeLineExcerpt` → `text.split(/(?<=\n)/)`, twice | 21.8 ms each, one string per line | at most 512 bytes from each end |
| `hash(text)` (sha256) | 5.5 ms | the content address — unavoidable |

Per-operation measurements, Node.js `v22.22.2`, macOS, on `160_000` lines of realistic log text.

`countLines` is the largest single item: `for..of` over a string iterates Unicode code points, which costs about 5x an indexed scan for an answer that only involves `\n`.

## Measurement

End to end through the real `context` handler, one 7.5 MiB observation, 20 projections after the first store:

```text
=====  main  =====
observation bytes = 7.5 MiB
20 projections = 2238 ms  (111.9 ms per provider request)

=====  this PR  =====
observation bytes = 7.5 MiB
20 projections =  594 ms  (29.7 ms per provider request)
```

3.8x, on the critical path in front of every model call, for a mechanism whose purpose is to make long trajectories cheaper.

## Change

No behaviour change, no API change, no new dependency:

- `containsReducerReceipt` scans with `indexOf` and checks the two boundary characters, instead of materialising every line;
- `countLines` uses an indexed `charCodeAt` loop (`countBufferLines` already reads bytes and stays as it is);
- `completeLineExcerpt` walks line breaks inward from the chosen end with `indexOf` / `lastIndexOf` and slices only the region it keeps.

## Equivalence evidence

A differential run of the old and new implementations over 60 000 randomly assembled payloads — empty strings, bare newlines, CRLF, astral characters, the receipt prefix in every position (own line, mid-line, at EOF, with a suffix) — across 12 excerpt variants (budgets 0/1/2/5/12/512 x head/tail):

```text
60000 random payloads x 12 excerpt variants: IDENTICAL
```

## Tests

Two added to `tests/observation-pack.test.ts`, pinning the semantics the rewrite must not lose:

- a payload that names `sol_pi_evidence_receipt_v1` **inside** a line (a `grep` command line, and the prefix with a trailing suffix) is still packed — the exemption is for a receipt, not for any line that mentions one;
- a payload whose lines each begin with an astral character (`𝄞`, one surrogate pair) reports the same `original_lines` a code-point scan produced.

These are characterization tests: they pass before and after by design, because this PR changes cost rather than behaviour. The evidence that behaviour is unchanged is the differential run above; the evidence that cost changed is the benchmark.

With the change, on the pinned Pi 0.84.2:

```text
```


gaoanze888 (approved):
Reviewed the allocation reductions against the previous semantics for line counts, exact receipt-line detection, and head/tail excerpts. The characterization coverage is appropriate and `npm run check` passes locally: 18 files / 141 tests, typecheck, and package dry-run. No blocking issue found.

Owen718:

gaoanze888 (approved):
Reviewed at exact head 120110082b0e.

The perf rewrite preserves semantics: `countLines` via `charCodeAt` agrees with code-point counting (the astral-character guard passes, and \n is a single UTF-16 unit so no surrogate false-positives); `containsReducerReceipt` now requires the prefix to be a whole line; and `completeLineExcerpt`'s backward walk uses `lastIndexOf("\n", start - 2)` to correctly skip the line-terminating newline (avoiding an empty-line loop).

Verification: focused `tests/observation-pack.test.ts` 15 passed, `tsc --noEmit` clean. Approving.

PR #34, opened 2026-09-13
fix(observation-pack): obs_recall decodes a partial character at a caller-chosen offset

### LINKED ISSUE #44: obs_recall decodes a partial character at a caller-chosen offset
## Summary

`obs_recall` returns replacement characters when the requested offset lands inside a multi-byte character, and reports an offset that does not match the bytes it returned.

`readRecallChunk()` trims the **end** of a chunk to a UTF-8 boundary (`trimUtf8End`) but decodes the **start** wherever the caller pointed.

https://github.com/NVlabs/SoL-Pi/blob/d7ecfc089944f0d04b80122a0a9a6ca0d786f3d0/src/sol-pi/extensions/observation-pack/observation.ts#L213-L248

## Reproduction

An archived observation beginning `☾☾☾ moon log` (each `☾` is three UTF-8 bytes), recalled at offset 1:

```js
await pi.tool("obs_recall").execute("recall-1", { id, offset: 1 }, undefined, undefined, ctx);
```

## Actual behavior

```text
chunkBytes= 15872 chunkLines=  88
chunk starts with U+FFFD: "\uFFFD\uFFFD☾☾" ...
```

and the header reads:

```text
[obs_recall id=obs_… offset=1 next_offset=15873 eof=false]
```

`offset=1` is not where the returned text starts, so a model that trusts the header cannot reconstruct the byte range it actually received.

## Impact

The tool's schema accepts any `offset: Type.Integer({ minimum: 0 })`, so this is reachable whenever the model picks an offset itself instead of following `next_offset` — seeking near the `original_bytes` reported in the placeholder, resuming from a value it recorded earlier, or bisecting a large observation.

Paging from offset 0 and following `next_offset` is unaffected, because every `next_offset` is already aligned.

## Expected behavior

The chunk should start at the next character boundary, and the reported offset should be the aligned one, so `offset` and `next_offset` continue to describe the bytes that were actually returned. The ledger can keep the requested value alongside it.

I have a focused fix with a regression test ready.

kaluli123123:
Fixes #44

## Summary

`obs_recall` returns replacement characters when the requested offset lands inside a multi-byte character, and reports an offset that does not match the bytes it returned.

`readRecallChunk()` trims the **end** of a chunk to a UTF-8 boundary (`trimUtf8End`) but decodes the **start** wherever the caller pointed.

## Reproduction

`main` @ `d7ecfc0`, Node.js `v22.22.2`. An archived observation beginning `☾☾☾ moon log` (each `☾` is three UTF-8 bytes), recalled at offset 1:

```text
chunkBytes= 15872 chunkLines=  88
chunk starts with U+FFFD: "��☾☾" ...
```

The tool's schema accepts any `offset: Type.Integer({ minimum: 0 })`, so this is reachable whenever the model picks an offset itself instead of following `next_offset` — seeking near `original_bytes` reported in the placeholder, resuming from a value it recorded earlier, or bisecting a large observation.

The header compounds it:

```text
[obs_recall id=obs_… offset=1 next_offset=15873 eof=false]
```

`offset=1` is not where the returned text starts, so a model that trusts the header cannot reconstruct the byte range it actually received.

## Change

`readRecallChunk()` advances past leading continuation bytes (at most three) and returns the aligned start as `RecallChunk.offset`. The tool reports that value in the header and in `details`; the ledger records both, as `requestedOffset` and `offset`.

`next_offset` and `eof` are unchanged in meaning, and paging from offset 0 with the returned `next_offset` — the documented path, already covered by `returns the exact original bytes across paged recall` — produces byte-identical output to before, because every `next_offset` is already aligned.

## Tests

One added to `tests/observation-pack.test.ts`: recalling the `☾☾☾ moon log` observation at offsets 1 and 2 returns text with no `U+FFFD`, starting at the second `☾`, and reports `offset=3` in both the header and `details`.

```text
AssertionError: expected '[obs_recall id=obs_54e14edd7d2523b472…' not to contain '�'
Tests  1 failed | 14 passed (15)
```

```text
npx vitest run            Test Files 18 passed (18) | Tests 140 passed (140)
```


gaoanze888 (approved):
Reviewed caller-selected continuation-byte offsets, aligned offset reporting in header/details/ledger, and unchanged exact paging via returned next_offset. `npm run check` passes locally: 18 files / 140 tests, typecheck, and package dry-run. No blocking issue found.

Owen718:

gaoanze888 (approved):
Reviewed at exact head 332bc4de6e5f.

The UTF-8 boundary fix is correct: leading continuation bytes are skipped to the next character boundary, `nextOffset = offset + end` keeps paging consistent, and `trimUtf8End` is bounded below by `start`. The returned `offset = offset + start` and the index.ts header/ledger updates line up with what the text actually starts at.

Verification: focused `tests/observation-pack.test.ts` 15 passed, `tsc --noEmit` clean. Mutation check on main's `readRecallChunk` reproduces U+FFFD (`��☾☾`) for the new test, confirming the regression guard is real.

Small heads-up (non-blocking): #33 also edits `observation.ts` in a different region; whichever lands second will need a trivial rebase.

PR #35, opened 2026-09-13
fix(action-fusion): then_run skipped for Git Bash/MSYS/Cygwin/WSL drive paths on Windows

### LINKED ISSUE #45: Action Fusion skips then_run for Git Bash/MSYS/Cygwin/WSL drive paths on Windows
## Summary

On Windows, a fused `edit`/`write` whose path comes from a Git Bash, MSYS, Cygwin, or WSL shell mutates one file while Action Fusion's queue and hash guard address another, so `then_run` is skipped with ENOENT.

`resolveToolPath()` handles `@`, `file://`, `~`, and `~/`, then calls `path.resolve()`.

https://github.com/NVlabs/SoL-Pi/blob/d7ecfc089944f0d04b80122a0a9a6ca0d786f3d0/src/sol-pi/extensions/action-fusion/file-queue.ts#L13-L24

Pi's `resolveToCwd()` does two more things on Windows, in `utils/paths.js`:

```js
if (process.platform === "win32") {
    normalized = normalizeWindowsShellPath(normalized);   // /c/src -> C:\src
}
if (options.expandTilde ?? true) {
    if (normalized.startsWith("~/") || (process.platform === "win32" && normalized.startsWith("~\\"))) {
        return join(home, normalized.slice(2));
    }
}
```

```js
export function normalizeWindowsShellPath(filePath) {
    if (!filePath.startsWith("/") || filePath.startsWith("//") || filePath.includes("\\")) return filePath;
    const match = filePath.match(/^\/(?:mnt\/|cygdrive\/)?([a-z])(?:\/(.*))?$/i);
    if (!match) return filePath;
    const suffix = match[2]?.replaceAll("/", "\\");
    return `${match[1].toUpperCase()}:\\${suffix ?? ""}`;
}
```

## Actual behavior

For `path: "/c/src/app.ts"` the built-in mutation writes `C:\src\app.ts`, while `resolveToolPath()` returns a literal `\c\src\app.ts` for the queue slot and the `assertUnchangedBeforeCommand()` hash. The mutation succeeds, the guard reads a path that does not exist, and the model gets:

```text
Error: [then_run:skipped] ENOENT: no such file or directory, open '...\c\src\app.ts'; the command was not run.
```

The same applies to `~\notes.txt`, which Pi expands and SoL-Pi treats as a relative filename.

## Scope of the claim

I do not have a Windows machine, so this is derived from Pi 0.84.2's own resolver source quoted above rather than executed on Windows. The divergence itself is checkable without a Windows runner; a confirmation on one would be welcome before merging anything.

This is the same parity class as #2 and #7, on a different input class — the file-URL fix in #4 and the Unicode-space work in #7 do not touch shell drive paths. Given #16, Windows users are running this.

## Expected behavior

Action Fusion should apply the same drive-path conversion and `~\` expansion Pi's own resolver applies, so the queue and hash guard observe the file the built-in mutation tool wrote. Both should stay no-ops off Windows, so a POSIX user whose repository genuinely contains `/c/src/app.ts` is unaffected.

I have a fix that mirrors Pi's function, with tests that drive the platform branch explicitly.

kaluli123123:
Fixes #45

## Summary

On Windows, a fused `edit`/`write` whose path comes from a Git Bash, MSYS, Cygwin, or WSL shell mutates one file while Action Fusion's queue and hash guard address another, so `then_run` is skipped with ENOENT.

`resolveToolPath()` handles `@`, `file://`, `~`, and `~/`, then calls `path.resolve()`. Pi's `resolveToCwd()` does two more things on Windows, in `utils/paths.js`:

```js
if (process.platform === "win32") {
    normalized = normalizeWindowsShellPath(normalized);   // /c/src -> C:\src
}
if (options.expandTilde ?? true) {
    if (normalized.startsWith("~/") || (process.platform === "win32" && normalized.startsWith("~\\"))) {
        return join(home, normalized.slice(2));
    }
}
```

```js
export function normalizeWindowsShellPath(filePath) {
    if (!filePath.startsWith("/") || filePath.startsWith("//") || filePath.includes("\\")) return filePath;
    const match = filePath.match(/^\/(?:mnt\/|cygdrive\/)?([a-z])(?:\/(.*))?$/i);
    if (!match) return filePath;
    const suffix = match[2]?.replaceAll("/", "\\");
    return `${match[1].toUpperCase()}:\\${suffix ?? ""}`;
}
```

So for `path: "/c/src/app.ts"` the built-in mutation writes `C:\src\app.ts`, while `resolveToolPath()` returns a literal `\c\src\app.ts` for the queue slot and the `assertUnchangedBeforeCommand()` hash. The mutation succeeds, the guard reads a path that does not exist, and the model gets:

```text
Error: [then_run:skipped] ENOENT: no such file or directory, open '...\c\src\app.ts'; the command was not run.
```

This is the same parity class as #2 and #7, on a different input class — the file-URL fix in #4 and the Unicode-space work in #7 do not touch shell drive paths. Given #16, Windows users are running this.

## Scope of the claim

I do not have a Windows machine, so this is reasoned from Pi 0.84.2's own resolver source rather than executed on Windows, and the tests below drive the platform branch with a stubbed `process.platform`. Happy to hold this until someone can confirm on a real Windows runner — the upstream code path is quoted above so the divergence itself is checkable without one.

## Change

`normalizeWindowsShellPath()` mirrors Pi's function exactly, including the `//`-prefixed UNC and already-native (`\`-containing) exclusions, and the tilde branch accepts `~\` on Windows. Both are no-ops off Windows, so a POSIX user whose repository genuinely contains `/c/src/app.ts` is unaffected.

`docs/compatibility.md` gains a sentence in the Action Fusion section next to the existing file-URL note.

## Tests

Three added to `tests/action-fusion-paths.test.ts`, using a `withPlatform()` helper that swaps `process.platform` around one call and restores it:

- under `win32`: `/c/src/app.ts` -> `C:\src\app.ts`, `/mnt/d/work/notes.md` -> `D:\work\notes.md`, `/cygdrive/e/x/y` -> `E:\x\y`, `/c` -> `C:\`, and `/usr/local/bin/pi`, `//server/share/file.txt`, `C:\already\native.ts` left alone;
- under `linux`: `/c/src/app.ts` unchanged, through both the helper and `resolveToolPath()`;
- `~\notes.txt` expands to the home directory under `win32` and stays an ordinary relative filename under `linux`.

```text
TypeError: normalizeWindowsShellPath is not a function
TypeError: normalizeWindowsShellPath is not a function
AssertionError: expected '/work/~\notes.txt' to be '/Users/…/notes.txt'
Tests  3 failed | 10 passed (13)
```

With the fix, on the pinned Pi 0.84.2 (macOS):

```text
npx vitest run            Test Files 18 passed (18) | Tests 142 passed (142)
```


gaoanze888 (approved):
Compared the converter and tilde branch with Pi 0.84.2 path normalization; the pure Windows path semantics match, and the POSIX behavior remains unchanged. `npm run check` passes locally: 18 files / 142 tests, typecheck, and package dry-run. No blocking correctness issue found. This overlaps #9, so merge resolution must preserve the order Unicode normalization → @ stripping → Windows drive normalization → file URL/tilde/cwd resolution; a real Windows run would still be valuable.

gaoanze888:
Integration follow-up: I merged this exact head after #9 in a disposable tree that also included the currently approved #6, #8, #13, #14, #15, #17, #23, #26, #30, #31, #33, and #34.

The expected conflict is limited to `file-queue.ts`. Resolving it as

```ts
normalizeWindowsShellPath(normalizeToolPath(filePath))
```

preserves the intended sequence: Unicode-space normalization → optional `@` stripping → Windows drive normalization → `file://` / tilde / cwd resolution. With the other complementary conflicts resolved similarly, the combined tree passes `npm run check`: **21 files / 223 tests**, typecheck, and package dry-run.

So this remains compatible with #9; it just needs that explicit single-file resolution if #9 lands first. This is still a macOS integration result, not a substitute for a native Windows run.

gaoanze888:
Merge-order note: this head conflicts with approved #52 in `file-queue.ts`. The integrated order must retain #52’s missing/non-string path guard before any string operation, then strip the optional `@`, then apply this PR’s Windows shell-path normalization, followed by file-URL/tilde/cwd resolution. Recommended order: #52 first, then rebase this branch and rerun Action Fusion tests.

lyttttt3333:
@HaozheLiu-ST @kaluli123123 Following up on the validation request in #45: we tested this PR on **native Windows Server 2022 with Git Bash**, with Ubuntu 22.04 as a control, using Node 22.23.2 and **both Pi 0.84.2 and 0.85.1**. Results were identical across the two Pi versions.

We compared main `2b791687` against the exact source/test patch from this PR at `7d7f082e`, and verified that the resulting runtime file matched the PR head. Pi itself was not modified.

### What is confirmed working

The fix aligns the Action Fusion queue/guard path with the path used by Pi's built-in mutation tools. Previously, Pi successfully modified the intended file, but the guard looked at a different path and skipped `then_run` with ENOENT.

Our independent probe exercised real `write`/`edit` operations and real Bash follow-up commands that checked the file content and wrote a sentinel. No filesystem/shell mocks, platform stubbing, or model calls were used in this probe.

- **Before the patch:** 12 Windows control cases succeeded; 16 affected cases reproduced the defect: mutation succeeded, then `[then_run:skipped]` + ENOENT, and the command did not run.
- **After the patch: 28/28 Windows cases completed both mutation and command successfully.** Coverage includes Git Bash `/c/...`, WSL-style `/mnt/c/...`, Cygwin-style `/cygdrive/c/...`, and `~\...`, with/without `@`, for both write and edit, plus native/relative/file-URL/home-slash controls.
- Linux independent checks passed **17/17 before and after**.

The WSL/Cygwin cases use those path strings with Windows-hosted Pi; we did not launch WSL/Cygwin runtimes or provision a UNC share.

### Remaining issue in this PR's new tests

`tests/action-fusion-paths.test.ts:81`, **"leaves a POSIX path alone off Windows"**, fails on native Windows. Changing `process.platform` to `linux` does not change the already-imported native `node:path.resolve` implementation. The assertion expects `/c/src/app.ts`, but gets `D:\c\src\app.ts` on this runner.

This is a portability problem in the new test, not a failure of the runtime fix in the scenarios above. Could you adjust this assertion to account for the host path implementation, or run the POSIX resolution assertion only on a real POSIX host? The normalization-helper assertion can remain platform-stubbed.

### Full-suite results

Counts below are per Pi version:

| Environment | Before | After |
| --- | --- | --- |
| Windows | 134 passed, 3 failed, 3 skipped (140 total) | 136 passed, 4 failed, 3 skipped (143 total) |
| Linux | 140/140 passed | 143/143 passed |

The original 140 tests retain the same outcomes. Of the 3 tests added by this PR, 2 pass and 1 fails on Windows; all 3 pass on Linux. We did not separately run the new tests against the old implementation; the before/after reproduction evidence comes from the independent probe.

The other 3 Windows failures are present **both before and after** this patch and should be tracked separately:

- `tests/evidence-preserving-reducer.test.ts:296`: POSIX mode assertion expects `0600`, but Windows reports `0666`; this is not an EPR reduction failure and does not establish the actual Windows ACL policy.
- `tests/pi-package-integration.test.ts:79`: the extension lookup uses a forward-slash path suffix and fails to match the Windows path; the later session prompt is not reached.
- `tests/observation-pack.test.ts:258`: recall of a symlink-replaced object resolves instead of rejecting with `ELOOP`. This warrants separate runtime/protection investigation, not dismissal as an assertion-only issue.

Type checking, public API checks, package dry runs, and `npm audit --audit-level=high` passed in all four environments.

Evidence and downloadable logs: [corrected native validation run](https://github.com/lyttttt3333/SoL-Pi/actions/runs/35078092838). The workflow remains red because full-suite failures are retained. An earlier run had an inline-command quoting error in our own probe; we corrected that harness issue, and the results above are from the corrected run.

**Bottom line: the intended runtime fix is independently validated on Windows for both Pi versions; the new cross-platform test needs adjustment before calling the Windows suite clean.**

gaoanze888 (changes requested):
The runtime fix is now independently validated on native Windows and its normalization matches Pi, but the exact head still adds a test that fails on the supported Windows platform.\n\n`tests/action-fusion-paths.test.ts` stubs `process.platform` to `linux` and then expects `resolveToolPath("/work", "/c/src/app.ts")` to use POSIX resolution. Changing `process.platform` does not change the host-specific implementation already exported by `node:path`; on native Windows that remains `path.win32.resolve`, and the independent matrix observed `D:\\c\\src\\app.ts` rather than `/c/src/app.ts`. Consequently both Windows jobs remain red even though the runtime probe passes 28/28 after the patch.\n\nPlease keep the pure `normalizeWindowsShellPath` branch assertion platform-stubbed, but gate the POSIX `resolveToolPath` assertion on an actual POSIX host (or refactor resolution to accept explicit path semantics for testing). The `~\\` stub similarly exercises the conditional branch on the host path implementation, not a complete Windows emulation; the native probe is the authoritative evidence there.\n\nLocal macOS `npm run check` passes: typecheck, 18 files / 142 tests, and package dry-run. Runtime behavior looks correct; this request is specifically to make the committed cross-platform suite truthful and green on Windows.

lyttttt3333:
@HaozheLiu-ST Revalidated the updated head `788d8c1e260a4ce8a3b1a06a542a3b3b522d3654` on native Windows Server 2022 with Git Bash and Ubuntu 22.04, using both Pi 0.84.2 and 0.85.1. Results are identical for both Pi versions.

**The previously reported Windows test portability failure is resolved.** The runtime fix remains successful in our independent real-file/real-shell checks.

| Check | Windows | Linux |
| --- | --- | --- |
| Action Fusion path tests | 13 passed, 0 failed, 1 skipped | 14/14 passed |
| Independent runtime checks | 28/28 passed | 17/17 passed |
| Full suite | 137 passed, 3 failed, 4 skipped | 144/144 passed |

The newly skipped Windows case is only the POSIX-host path-resolution assertion; it executes and passes on Linux. The other three PR-added path tests pass on Windows. Windows path behavior is also exercised by the 28 real mutation/follow-up cases, so this is not a substitute for testing the Windows functionality.

All original 140 tests retain their baseline outcomes. The 3 remaining Windows failures are the same pre-existing EPR permission assertion, ObservationPack symlink-recall protection issue, and package-integration path lookup issue. This PR no longer adds a Windows failure; the overall workflow remains red only because those existing suite failures are retained.

Typecheck, public API checks, package dry runs, and the high-severity audit gate also passed in all four environments. No model calls or upstream Pi modifications were involved. WSL/Cygwin-style path strings were tested with Windows-hosted Pi, not native WSL/Cygwin runtimes.

Evidence: https://github.com/lyttttt3333/SoL-Pi/actions/runs/35241648976

Owen718:

lyttttt3333:
@HaozheLiu-ST We revalidated the current head on native Windows Server 2022 with Git Bash and Ubuntu 22.04, using both Pi 0.84.2 and 0.85.1. The runtime fix passed our independent real-file / real-shell checks, including the Windows shell-path cases this PR targets.

This is a small, focused Action Fusion path-parity fix. The remaining Windows full-suite failures are pre-existing and unrelated to this PR. From the validation side, this looks ready to merge. If merged together with #9, please preserve the resolver order: Unicode-space normalization -> optional @ stripping -> Windows drive-path normalization -> file URL / tilde / cwd resolution.

HaozheLiu-ST (approved):
LGTM. Thanks @lyttttt3333 for the native Windows/Linux validation. I resolved the overlap with #9 while preserving Unicode-space normalization -> optional @ stripping -> Windows drive normalization -> file URL / tilde / cwd resolution, and added four composition regressions. The updated tree exactly matches the combined #9/#35/#61 tree validated on macOS with Pi 0.85.1: 158 tests, typecheck, package inspection, public API checks, all-mechanisms checks, and the high-severity audit gate passed.

PR #47, opened 2026-09-14
fix(reducer): preserve distinct nonblank lines in failed output

hi-neason:
## Problem and change

A failed command can produce `0 errors reported` followed by `fatal: missing symbol x`. A receipt quoting only the first line must not replace the original output, even if the model labels that quote `failure` or `fatal`.

For failed tool results, validation now requires every distinct nonblank source line to appear complete in the verified quotes, including indentation. It does not infer failure semantics from keywords or evidence labels. Missing lines, empty source content, or `uncertain=true` cause fallback to the original output. The reducer instructions describe the same rule, and the unused failure-keyword heuristic is removed. Successful receipt validation is unchanged.

## Trade-off

Failed-log reduction mainly benefits repetitive output: identical lines can be retained once. Logs whose required lines cannot fit the evidence limits fall back to the original output. This preserves distinct nonblank line content, but does not guarantee original ordering, repetition counts, or blank lines. Evidence coverage does not prove the semantic accuracy of the model-assigned labels. Validation still follows the reducer call, so a rejected receipt may incur a call without reducing output. No measured token-saving claim is made.

## Validation

Regression coverage includes benign and negated failure keywords, unknown and Chinese diagnostics, omitted multiline details, shortened or fragmented lines, indentation, overlong lines, uncertain receipts, empty source content, LF/CRLF, and successful repeated-line reduction across evidence kinds.

With Node 24.15.0 and the branch's Pi 0.84.2 dependencies:

- Targeted reducer and all-mechanisms tests: **50 passed**.
- `npm run check`: **170 tests passed across 18 files**, typecheck and package dry-run passed.
- `node scripts/check-pi-compat.mjs` and `git diff --check`: passed.

The same four changed files were also checked in an isolated upstream source snapshot with Pi 0.85.1: **171 tests passed across 19 files**, plus typecheck, package dry-run, and Pi API compatibility checks. These are local validation results, not GitHub CI results. No lint script or ESLint dependency is configured.

## Compatibility

Remains a standalone extension using Pi's public APIs. No Pi core changes, configuration changes, or new dependencies.


gaoanze888 (changes requested):
Reviewed exact head `98f026886ee01f3e790afd32ea7f140d9591a799`. The fail-open plumbing and the new benign-label tests are sound, and locally `npm run check` passed (18 files / 141 tests), as did `node scripts/check-pi-compat.mjs` and `git diff --check`.

There is still a concrete bypass of the intended evidence guarantee. `FAILURE_SIGNAL` is an unbounded substring heuristic, and the new condition accepts any failure/fatal quote that matches it, including negated success text. I reproduced `validateReceipt()` returning `ok: true` for a failing source containing `0 errors reported\nfatal: missing symbol x` when the only receipt evidence is `{ "kind": "failure", "quote": "0 errors reported" }`. The actual fatal evidence is still omitted. Similar cases include `no failures`, `timeout disabled`, and `assertions passed`.

Please add a regression for this negated/success-context case and tighten the criterion so that such a quote cannot satisfy the required failure evidence. The conservative option is preferable here: if the heuristic cannot establish that retained evidence represents the observed failure, preserve the original output rather than claiming a reduced failure receipt.

hi-neason:
In response to [the review](https://github.com/NVlabs/SoL-Pi/pull/47#pullrequestreview-5205942280):

Addressed in 2565836. Failed receipts now must retain every distinct nonblank source line, including indentation, regardless of keywords or evidence labels. Missing content or `uncertain=true` triggers fallback to the original output.

Regression tests cover your `0 errors reported` example, other benign/negated phrases, and unfamiliar diagnostics. All 170 local tests pass; validation with Pi 0.85.1 also passes (171 tests).

The trade-off is more conservative reduction: failing logs mainly benefit from repeated-line deduplication. Could you take another look?

gaoanze888 (approved):
Re-reviewed exact head `25658369a1885c6a36af237004f2bcd99d968971`. The previous negated-keyword bypass is closed by the new content-based rule: every distinct nonblank source line in a failed result must appear complete in verified quotes, independent of model-assigned kind or keyword semantics; uncertain or unrepresentable failures fall back to the original output. I checked partial/combined lines, indentation, LF/CRLF, repeated lines across evidence kinds, and unfamiliar diagnostics. Locally, `npm ci --ignore-scripts`, `npm run check` (18 files / 170 tests), `node scripts/check-pi-compat.mjs`, and `git diff --check` passed. The conservative reduction rate is an explicit trade-off and preserves correctness. No blocking issues found.

gaoanze888:
Merge-order note from an approved-head integration run: this receipt-hardening head is the right semantic baseline, but applying current #8 or #26 afterward leaves their failure-receipt test fixtures quoting only part of the source, so the combined suites fail (10 fixtures for #8, 1 for #26). This is fixture drift rather than a production-logic conflict. Recommended order: merge #47 first, then rebase #8/#26 and update those receipts to cover every distinct nonblank failed-output line before merging.

Owen718:

Issue #20, opened 2026-09-11
then_run fusion silently does nothing on Pi 0.85.1 (works on pinned 0.84.2)


AdrianTJ:
Hi. Action Fusion's `then_run` follow-up never executes on Pi 0.85.1. It fails silently. Neither an error nor a marker appears.

Repro: with `actionFusion: true`, call `edit` on any file with `then_run: {"command": "cat <same-file>", "timeout": 15}`. The edit applies. The result is Pi's plain success message. The command never runs, so the agent re-runs the check itself.

Environment: Pi 0.85.1, SoL-Pi 0.1.0, Node 26, macOS. Config is fusion plus ObservationPack on, nothing else.

Key data point: the identical test under your pinned Pi 0.84.2 (isolated via npx, SoL-Pi-only profile) returns the fused observation correctly, with `[then_run:succeeded]` plus the command output in a single call. The mechanism is sound. The handoff breaks somewhere between 0.84.2 and 0.85.1.

What I ruled out: registration is fine. Pi's registry builds builtins first, then extension tools overwrite by name, so the replacement genuinely owns the `edit` slot on this version. I also read `executeMutationThenRun` end to end. Every path with `thenRun` defined yields markers or appended output. The only bare-return path is `thenRun === undefined`. So the argument looks dropped between the model call and execution.

Thanks for the rigorous project. The held-out methodology in the blog is why I trusted this enough to dig in rather than uninstall.

AdrianTJ:
Related but distinct from #7, noting here to save triage: I checked whether this could be the same path-parity failure. I don't think it can be, because all paths in my repro are pure ASCII (/tmp/solvf2/cfg.txt), so there is nothing for Unicode normalization to diverge on. Additionally, #7 fails loudly with [then_run:skipped], while my failing case returns Pi's plain success message with zero markers of any kind. Separate failure mode, same then_run surface.

xqicxx:
I could not reproduce this on Pi 0.85.1 — but I found the mechanism, and it is not a 0.85.1 regression.

**Setup**: Pi 0.85.1, SoL-Pi 0.1.0 @ `d7ecfc0`, macOS Apple Silicon, `deepseek-v4.1-flash` @ max thinking, isolated
`PI_CODING_AGENT_DIR`, `actionFusion: true`, `observationPack: false`. Same prompt in all three arms: *"add a mean(values)
function to calc.js and run it to verify"* — `then_run` is never mentioned in the prompt.

[...]

PR #49, opened 2026-09-14
fix: load and fuse correctly on Pi-compatible hosts with a reduced export surface

### LINKED ISSUE #48: SoL-Pi cannot load, and Action Fusion publishes a parameterless tool, on Pi-compatible hosts with a reduced export surface
Two independent failures when SoL-Pi is loaded by a Pi-compatible host that is not the official `pi` binary. I hit both on omp (oh-my-pi) 18.1.19, a Pi fork that re-exports `@earendil-works/pi-coding-agent` from its own bundle. Both are host-surface assumptions in SoL-Pi rather than host bugs, and both are distinct from #20/#22 (there the fused tool is registered but the model never calls it; here the fused tool is registered and *is* called).

Environment: omp 18.1.19 (Homebrew, single compiled binary), SoL-Pi 0.1.0 @ `d7ecfc0`, macOS arm64, Node 24 / Bun. Config: `actionFusion: true`, `observationPack: true`, other two false.

## 1. Install fails: two context APIs are imported as required named bindings

```
$ omp install git:github.com/NVlabs/SoL-Pi
✘ Failed to install: Plugin sol-pi extension validation failed:
  .../sol-pi/src/sol-pi/index.ts: Failed to load extension:
  Export named 'findCutPoint' not found in module '<host>:@oh-my-pi/pi-coding-agent'.
```

`online-context-compact/extension.ts` imports `findCutPoint` and `sessionEntryToContextMessages` statically. This host exports everything else SoL-Pi uses (`buildSessionContext`, `estimateTokens`, `createEdit/Write/BashToolDefinition`, `getAgentDir`, `CONFIG_DIR_NAME`, `pi-tui` `Container`/`Text`, `pi-ai/compat#complete`) but not those two — they were reworked out of its public surface.

Impact is larger than the mechanism: the failing import aborts module evaluation, so the package never loads and all four mechanisms are unavailable, even with `onlineContextCompact: false`. Both symbols are used only inside `nativeCompactionFeasible()`.

Related to #11, but the failure here is a missing export inside an accepted version, not an out-of-range version.

## 2. Action Fusion publishes a fused tool with no built-in parameters

`createActionFusionExtension()` rebuilds the fused schema from `template.parameters.properties`. On Pi that is a TypeBox `Type.Object` and the read is exact. This host describes tool parameters with its own schema value, where:

- `typeof parameters === "function"` (the schema is callable, not a plain object),
- `parameters.properties` is `undefined`,
- the contract is reachable only via `parameters.toJsonSchema()` → `{"type":"object","properties":{"path":{...},"content":{...}},"required":["path","content"]}`.

So `{...template.parameters.properties, then_run}` collapses to `{then_run}`, and the published `write` tool no longer declares `path` or `content`. Two observable outcomes, depending on how strictly the host filters arguments against the published schema:

- arguments filtered → `resolveToolPath(ctx.cwd, input.path)` receives `undefined`:
  `undefined is not an object (evaluating 'filePath.startsWith')` — every fused call throws;
- arguments forwarded → the mutation succeeds and `then_run` is dropped with no marker, the silent shape described in #20.

Verified on the host: before the fix, a `write` call carrying `then_run` returned only the plain success text; after the fix the same call returns

```
[fused.txt#BF2C]
Successfully wrote 5 bytes to fused.txt
[then_run:succeeded]
5 fused.txt
```

A second, smaller point in the same area: this host's `edit` tool takes a single multi-file patch string rather than `path` + `edits`, so a fused `edit` has no single target to key the file queue and the pre-command hash check on. Replacing it produces a tool that cannot work regardless of schema handling.

## Suggested fix

Both are one-line-of-principle changes that are no-ops on Pi:

1. resolve the two context APIs through the module namespace and fall back locally, so a reduced surface costs a pre-check refinement instead of the whole package;
2. read the built-in properties through a helper that prefers `parameters.properties` and otherwise rebuilds them from `parameters.toJsonSchema()` (preserving optionality from `required`), and skip replacing a mutation tool whose contract has no `path`.

PR incoming with both, plus a test that covers the rebuild and fails without it. Existing suite stays green (143 tests).

luw2007:
Fixes #48.

Two host-surface assumptions make SoL-Pi unusable on a Pi-compatible host that is not the official `pi` binary. Reproduced on omp (oh-my-pi) 18.1.19, a Pi fork that re-exports `@earendil-works/pi-coding-agent` from its own bundle. Both changes are no-ops on Pi.

## 1. `fix(action-fusion): rebuild fused tool parameters from the host schema`

Action Fusion re-declares the built-in edit/write parameters from `template.parameters.properties`. That is exact on Pi (TypeBox `Type.Object`). The host in #48 describes the same tools with a callable schema value that has no `properties` and exposes the contract only through `toJsonSchema()`, so the spread collapses to `{then_run}` and the fused `write` no longer declares `path` or `content`. Depending on how the host filters arguments, every fused call either throws (`undefined is not an object (evaluating 'filePath.startsWith')`, `input.path` undefined) or silently drops `then_run` — the symptom shape of #20 with a different cause.

`hostToolProperties()` prefers `parameters.properties` and otherwise rebuilds the properties from `toJsonSchema()`, keeping non-`required` keys optional. When the resulting edit/write contract has no `path`, the built-in tool is left in place instead of being replaced by a fused tool that cannot key its file queue and pre-command hash check on a single target (that host's `edit` takes a multi-file patch string).

## 2. `fix(online-context-compact): resolve optional context APIs at runtime`

`findCutPoint` and `sessionEntryToContextMessages` are static named imports, but they are used only by the `nativeCompactionFeasible()` pre-check. On a host that does not export them, the import aborts module evaluation, `pi install` fails validation, and all four mechanisms are unavailable — including with `onlineContextCompact: false`. They are now resolved through the module namespace with local fallbacks (message-count from the entry's own message; backwards token walk for the cut point, reporting a whole-entry cut since turn splitting is a host-side refinement). Pi keeps its exact current behavior because both functions are present there and take precedence.

## Efficiency rationale

No new token cost: both changes are load-time and schema-time only, and neither adds a model call or context content. They restore Action Fusion's existing saving (one model round-trip per fused mutation) on hosts where it currently cannot be reached at all, and they stop one unreachable mechanism from disabling the other three.

## Validation

- `npm run check` (typecheck + full suite + `npm pack --dry-run`): **143 tests, 19 files, all passing**, up from 139/18.
- New `tests/action-fusion-host-schema.test.ts` covers the TypeBox path, the `toJsonSchema()` rebuild, optionality preservation, and the no-surface case. Falsified by mutation: reverting the rebuild branch fails 2 of its 4 tests, and the rest of the suite still passes — so the existing tests do not cover this and the new ones do.
- End-to-end on the affected host, real session, one `write` call carrying `then_run`:

```
CALL   write {"path":"fused.txt","content":"FUSED","then_run":{"command":"wc -c fused.txt"}}
RESULT [{"text":"[fused.txt#BF2C]\nSuccessfully wrote 5 bytes to fused.txt"},
        {"text":"[then_run:succeeded]\n5 fused.txt"}]
```

Before the change the same call returned the plain success text with no marker; before change 2 the package did not load at all.

## Trade-offs

- The `toJsonSchema()` rebuild derives optionality from the `required` list only; a host that expresses optionality another way would publish stricter parameters than it declares. Pi is unaffected (first branch).
- The fallback cut point is coarser than Pi's `findCutPoint`: it never reports a split turn, so on a reduced host the feasibility pre-check can consider slightly less history compactable. It never over-reports.
- Skipping the fused `edit` on hosts without a `path` parameter means Action Fusion covers `write` only there. That is deliberate: the alternative is a tool that throws on every call.


gaoanze888 (changes requested):
Reviewed exact head `ddcdf9d26ff771aa0f7b61442bc26f82ee9668c7`. Locally, `npm ci --ignore-scripts`, `npm run check` (19 files / 143 tests), `node scripts/check-pi-compat.mjs`, and `git diff --check` pass. The reduced-host goal is valuable, but two schema-adaptation cases currently break the promised conservative behavior:

1. An ordinary JSON Schema object with top-level `properties` returns early before consulting `required`. I added a local dynamic regression using `{ properties: { path, encoding }, required: ["path"] }`; after `hostToolProperties()` plus `Type.Object()`, the result requires both `path` and `encoding`. A valid optional host argument is therefore made mandatory. Please normalize both direct-properties and `toJsonSchema()` shapes through the same required/optional conversion.

2. Introspection exceptions abort extension initialization. A `.properties` getter that throws propagates directly (dynamically reproduced), as do a throwing/null `toJsonSchema()` and boolean JSON Schema fragments passed to `Type.Optional`. On a compatibility shim this should conservatively return no usable properties and leave the host tool in place, not prevent Action Fusion / package loading. Please isolate introspection, validate the returned JSON Schema shape/fragments, and add regressions for throwing accessors/calls and non-object fragments.

The OCC fallback appears conservative for the feasibility pre-check; I did not find a blocking issue there.

Owen718:

PR #50, opened 2026-09-14
fix: harden ObservationPack for non-persistent sessions and cross-platform runs

atharvaHJoshi:
## Summary

Makes SoL-Pi robust across session types and platforms, and documents every fixed issue in the README.

### ObservationPack reliability
- **Fail-open on non-persistent sessions**: when Pi provides no persistent session directory, the context projection now returns the context unchanged (with a log) instead of throwing. Previously this crashed context handling.
- **Windows symlink protection**: `O_NOFOLLOW` is undefined on Windows. Replaced with portable `constants.O_NOFOLLOW ?? 0` plus explicit `lstat` checks and a synthesized `ELOOP` error for parity. Symlink-dependent tests are skipped on Windows.

### Cross-platform / tooling
- **npm >= 12 pack output**: `tests/package.test.ts` now handles both the old array and the new single-object pack report; uses `npm.cmd` on Windows.
- **CI matrix**: new `.github/workflows/ci.yml` runs `npm ci`, `npm run check`, `npm audit --audit-level=high`, and `check-pi-compat.mjs` on ubuntu, macOS, and Windows.
- **vitest**: bumped to `^4.1.11`; lockfile refreshed.

### Documentation
- **README**: added a comprehensive *Fixed Issues* section listing each issue, how it was found, and how it was solved, plus the discovery/resolution processes.
- **agents-install.md**: documented bash/PowerShell/cmd usage across platforms.

## Validation
- `npm run check` — typecheck + full test suite + package inspection: passed
- `npm audit --audit-level=high` — 0 vulnerabilities
- `npx vitest run tests/all-mechanisms.test.ts` — 4 passed


gaoanze888 (changes requested):
Reviewed exact head `ac2ac9f9e06cc60d3245a87208414f66fe69becb`. Locally, `npm ci --ignore-scripts`, `npm run check` (18 files / 140 tests), `node scripts/check-pi-compat.mjs`, and `git diff --check` pass. The npm 12 pack parser and basic CI matrix are useful, but two correctness issues remain:

1. On Windows, `O_NOFOLLOW` becomes `0`, so `lstat(path)` followed by a separate `open(path)` is still a check/use race. A pathname can be replaced by a symlink or reparse point between those calls; `open()` follows it, and `handle.stat()` then reports the target as a regular file. This can affect both recall and existing-object reuse. The two symlink tests are skipped on Windows, so the new matrix does not verify the claimed parity. Please use a race-safe Windows no-follow/open-reparse-point mechanism, or fail closed where that cannot be guaranteed; do not describe the current lstat sequence as closing the same holes.

2. The context handler catches every `runtimeRoot()` error as a non-persistent-session fallback. `runtimeRoot()` also rejects unsafe session IDs, so an invariant/security failure is now silently downgraded. Please narrow fail-open to the explicit missing-session-directory case (preferably with a typed/distinguishable error) and retain rejection for unsafe IDs, with tests for both.

Separately, the 124-line README fixed-issues history is much broader than this patch and contains questionable attribution (for example, the overrides already exist on the base). I recommend removing it from this focused fix.

Owen718:

PR #52, opened 2026-09-14
fix: reject missing write/edit path with a tool error instead of a TypeError

cexll:
## Summary

When a host shim delivers a `write`/`edit` `execute()` call without a `path` (possible when the host's merged tool schema does not enforce the required keys), Action Fusion's `resolveToolPath` crashed with a raw `TypeError`:

```
undefined is not an object (evaluating 'filePath.startsWith')
```

This guards the input in `resolveToolPath` and surfaces a clear tool-visible error instead:

```
write/edit tool input is invalid. `path` must be the target file path.
```

## Testing

- New regression test in `tests/action-fusion-paths.test.ts`: a write call missing `path` rejects with the guard error, performs no file mutation, and runs no `then_run` command.
- `bun test tests/action-fusion.test.ts tests/action-fusion-paths.test.ts` - 27 pass.

Reproduced against omp's legacy compatibility shim (`omp/18.1.21`, SoL-Pi 0.1.0 installed at `~/.omp/plugins`): before the guard, a path-less write call surfaced the raw `TypeError`; after, the model receives the actionable error and can retry with a valid `path`.


gaoanze888 (approved):
Reviewed exact head `980fef81782625e9c2b36f3562499a09911c79ce`. The guard executes before path normalization, queue acquisition, mutation, or `then_run`, and the regression drives the registered fused write tool while proving both mutation and command counts remain zero. Locally, `npm ci --ignore-scripts`, `npm run check` (18 files / 140 tests), `node scripts/check-pi-compat.mjs`, and `git diff --check` passed. This is a useful defensive boundary even though #49 remains the broader fix for reduced-schema hosts. I found no blocking correctness issues.

Owen718:

PR #57, opened 2026-09-15
fix: keep ObservationPack symlink checks portable

### LINKED ISSUE #16: Windows support: O_NOFOLLOW is undefined, so the ObservationPack symlink guard silently no-ops; 4 tests fail on Windows

kaluli123123:
## Summary

- keep ObservationPack symlink rejection effective when `O_NOFOLLOW` is unavailable
- verify every opened object against its current pathname before reading or writing bytes
- revalidate the ObservationPack storage directories after each open
- invoke `npm.cmd` for the package test on Windows

## Root cause

Node does not expose `fs.constants.O_NOFOLLOW` on Windows, so the original bitwise flag expression silently degraded to a plain open. A split pathname check was also insufficient because the path or an ancestor could be replaced between validation and open. The package test separately spawned `npm` instead of the Windows executable name.

## Changed behavior

ObservationPack still uses `O_NOFOLLOW` where available. After every object open, it now revalidates the `observation-pack` and `objects` directories, requires both the pathname and handle to describe regular files, and compares their device/inode identity before any content is read or written. This detects both persistent symlink substitution and a path restored to a different object after open. Platforms without `O_NOFOLLOW` retain a functional idempotent storage and recall lifecycle through the same post-open identity checks. Compatibility documentation records this boundary.

The package test selects `npm.cmd` on Windows.

## Test plan

- `npm run check` — 20 test files, 142 tests passed
- `npx vitest run tests/all-mechanisms.test.ts` — 4/4 passed
- `npm audit --audit-level=high` — exited 0; two existing moderate Vitest advisories remain
- `git diff --check` — passed

New deterministic regressions cover an ancestor-directory replacement, the exact open-then-path-restored boundary, and a complete lifecycle with `O_NOFOLLOW` unavailable.

## Issue reference

Fixes #16

## Incomplete gates

- Native Windows execution was not available in this environment; the no-`O_NOFOLLOW` branch is exercised through a mocked Node filesystem capability.
- Remote CI and maintainer review remain pending.


gaoanze888 (changes requested):
Reviewed exact head `64e3cdd986d9ad958166937b425282ec58835c90`. `npm run check` passes locally (18 files / 139 tests), Pi compatibility and diff-check pass, and the `npm.cmd` change is correct. However, the claimed portable symlink protection is not race-safe on Windows:

- `O_NOFOLLOW` becomes 0. Existing-object reuse does `lstat(path)` then separately `open(path)`; a replacement between calls is followed.
- Recall does `open(path)` first, then `lstat(path)`; a symlink can be followed at open and the pathname swapped back before lstat, while reads continue from the already-open target handle.

The static symlink test cannot cover either race and there is no Windows execution evidence. This is the same TOCTOU class identified on #50/#58. Please use a true Windows open-reparse/no-follow mechanism, or fail closed when atomic no-follow cannot be guaranteed; a split pathname check cannot be described as equivalent to POSIX `O_NOFOLLOW`.

kaluli123123:
Addressed in `a3e3164`.

Existing-object reuse and recall now fail closed when the platform cannot provide atomic no-follow opens. The split pathname checks were removed, so the implementation no longer describes `lstat()` plus `open()` as equivalent to POSIX `O_NOFOLLOW`. Platform-conditional regressions cover the supported atomic path and the unsupported fail-closed path.

Local validation passes: 139 tests, all-mechanisms checks, Pi compatibility, `git diff --check`, and the high-severity audit gate. Please re-review the current head.

gaoanze888 (changes requested):
Re-reviewed exact head `a3e316494ef1fd9aa42cd8a2247300d812ce4392`. Same-handle `O_NOFOLLOW` correctly protects the final object component, and unsupported platforms now fail closed on reuse/recall. Two blockers remain. First, on Windows the normal lifecycle cannot produce a placeholder after the first projection: the second `ensureStored()` hits EEXIST and fails because atomic no-follow is unavailable; existing unconditional projection/restart/paged-recall tests would therefore fail on Windows. Please either provide a safe Windows mechanism or make the documented fail-closed behavior and complete Windows test expectations consistent. Second, `lstat(objectsDir)` followed by `open(fullPath, O_NOFOLLOW)` does not protect ancestor components; O_NOFOLLOW applies only to the final component, so replacing `objectsDir` with a symlink between those calls can redirect creation outside the session root. Please bind traversal to trusted directory handles/platform APIs or otherwise eliminate/fail closed on that parent-directory race.

kaluli123123:
Addressed in `c879b71`.

Every opened ObservationPack object is now verified before any bytes are read or written: both storage directories are revalidated, the pathname and handle must describe regular files, and their device/inode identities must match. This detects both an ancestor-directory substitution and the exact open-then-path-restored race. Where Node does not expose `O_NOFOLLOW`, the same post-open identity checks preserve idempotent storage and recall instead of failing midway through the lifecycle. The compatibility documentation and PR description now state this boundary.

Validation passes: `npm run check` (20 files / 142 tests), Pi compatibility, all-mechanisms (4/4), `git diff --check`, and the high-severity audit gate. Deterministic regressions cover both replacement boundaries and a no-`O_NOFOLLOW` lifecycle. The audit still reports the existing two moderate Vitest advisories. Native Windows execution was not available, so that remains an explicit incomplete gate. Please re-review the current head.

gaoanze888 (changes requested):
The exact-head identity checks close the direct `objects` replacement races and restore the no-`O_NOFOLLOW` lifecycle; the focused suites pass 17/17. A higher ancestor remains unchecked, however. `verifyOpenedObject()` only lstats `observation-pack` and `objects`. If the session-derived runtime root (`<sessionDir>/sol-pi/<sessionId>`) is itself a symlink/junction, both checked directories and the file are ordinary objects under the external target, so pathname/handle dev+ino match and the write succeeds outside the session storage boundary. I reproduced this on `c879b7158678`: symlinking that runtime-root component to a temp directory made `ensureStored()` successfully write `sensitive observation bytes` under the external target. Please validate the complete chain from a trusted session root (and define/test Windows junction/reparse behavior), or keep the no-`O_NOFOLLOW` path fail-closed. The compatibility claim that storage directories are revalidated is currently too broad.

Owen718:

gaoanze888 (approved):
Re-reviewed at exact head eec41aaa95.

The runtime-root ancestor race I reproduced earlier is now closed: `verifyStorageDirectories` computes the subpath from the trusted `sessionDirectory` via `relative()`, and that subpath includes `sol-pi/<sessionId>`, so the session-id (runtime root) component is `lstat`'d and rejected if it is a symlink/junction — not just `observation-pack` and `objects`. `verifyOpenedObject` also re-checks the chain after open and compares pathname vs handle dev+ino, closing the path-restoration race. The doc now scopes the honest remaining limit (not atomic against a process that keeps replacing ancestors; native Windows reparse-point types unvalidated).

Non-blocking: there is no dedicated regression test pinning the runtime-root symlink case itself (the existing tests cover the objects-dir and object-file symlink cases, which exercise the same mechanism). Adding one would strengthen the guarantee but does not block merge. Approving.

PR #58, opened 2026-09-15
fix: reject symlinked reducer archive objects

### LINKED ISSUE #55: Evidence-Preserving Reducer accepts symlinked archive objects during reuse
## Summary

Evidence-Preserving Reducer can accept a symlink in place of an existing content-addressed archive object when the symlink target contains the expected bytes. This violates the archive integrity boundary and differs from the ObservationPack storage checks, which reject symlinked objects.

## Reproduction

On `main` at the current repository head, call `archiveBody()` once for a body, replace the resulting `<hash>.txt` object with a symlink to another regular file containing the same body, then call `archiveBody()` again with the same body.

The second call succeeds because `archiveBody()` handles `EEXIST` with `readFile(path, "utf8")`. `readFile()` follows the symlink, and the byte/hash comparison passes.

## Expected behavior

Content-addressed archive objects should be regular files, not symlinks. Reuse should reject a symlinked object even when its target has identical contents.

## Impact

A local archive path replacement is accepted as an authentic archive object. This weakens the reducer's documented evidence and storage-integrity boundary and can make later path-based evidence read through an attacker-controlled link.

## Suggested fix

Use `lstat()` before reusing an existing object and reject symbolic links or non-regular files. Add a regression test covering a same-content symlink.

## Verification boundary

A focused unit test should prove the existing object is rejected when replaced by a symlink. The full `npm run check` suite should remain green.

kaluli123123:
## Summary

- reject symlinked Evidence-Preserving Reducer archive objects during reuse
- add regression coverage for a same-content symlink replacement

## Root cause

When a content-addressed archive path already existed, the reducer used `readFile()` to verify it. `readFile()` follows symlinks, so a symlink pointing to a file with identical bytes passed the hash and content checks.

## Changed behavior

Existing archive objects must first be confirmed with `lstat()` to be regular non-symlink files. A symlink or other non-regular object is rejected before its contents are reused.

## Test plan

- `npm run check` — 18 test files, 140 tests passed
- `npx vitest run tests/all-mechanisms.test.ts` — 4/4 passed
- `git diff --check` — passed

## Issue reference

Fixes #55

## Incomplete gates

- `npm audit --audit-level=high` exits successfully but reports the existing two moderate Vitest advisories.
- CI and maintainer review remain pending.


gaoanze888 (changes requested):
Reviewed exact head `070b1da216dc05eb5fbac027899c11611c2cfaf9`. The static symlink regression is useful, and `npm run check`, `node scripts/check-pi-compat.mjs`, and `git diff --check` pass locally. However, the implementation still has a check/use race: `lstat(path)` validates one directory entry, then `readFile(path)` reopens the pathname and follows whatever is there at that later instant. A concurrent replacement between those calls can therefore still make archive reuse read through a symlink, so the claimed non-symlink integrity property is not enforced.

Please use the same object-level pattern already used by ObservationPack: open the existing object with `O_RDONLY | O_NOFOLLOW`, call `stat()` and `readFile()` on that one `FileHandle`, and close it in `finally`. That both rejects an initial symlink and pins subsequent verification to the opened inode instead of reopening the pathname. The existing static symlink test should remain; a deterministic test around the handle/path replacement boundary would further lock down the TOCTOU behavior.

kaluli123123:
Addressed in `61f1283`.

Archive reuse now opens the existing object with `O_RDONLY | O_NOFOLLOW`, verifies it with `FileHandle.stat()`, reads through that same pinned handle, and closes it in `finally`. Platforms without atomic no-follow support fail closed. The regression remains platform-aware while covering the handle-based path.

Local validation passes: 140 tests, all-mechanisms checks, Pi compatibility, `git diff --check`, and the high-severity audit gate. Please re-review the current head.

gaoanze888 (approved):
Re-reviewed exact head `61f12834fc64d65856ede68582a5caa3260cd85a`. Existing archive objects are now opened atomically with `O_NOFOLLOW`, then stat/read/verified through that same handle and closed in `finally`; no path-based lstat/read window remains. Platforms without atomic no-follow fail closed on reuse rather than falling back to a racy check. The focused and full suites pass (18 files / 140 tests). No blocking issue found; the explicit portability trade-off is that Windows cannot reuse an already-existing object.

Owen718:

PR #59, opened 2026-09-15
fix(compaction): align and bound plan string validation

### LINKED ISSUE #65: update_plan schema causes llama.cpp PEG grammar parse failure
### `update_plan` schema causes llama.cpp PEG grammar parse failure

When `onlineContextCompact` is enabled, SoL-Pi registers `update_plan`. With llama.cpp `peg-native` tool calling, this schema can exceed llama.cpp's grammar repetition limits.

Minimal repro:

```text
Pi + SoL-Pi
onlineContextCompact = true
llama.cpp backend
```

Run any task that enables tools. llama.cpp logs:

```text
Chat format: peg-native

parse: error parsing grammar:
number of repetitions exceeds sane defaults

E failed to parse grammar
```

The generated grammar contains large repetitions from `update_plan`, for example:

```text
tool-update-plan-arg-progress-schema-decisions-item ::= "\"" char{0,1000} "\"" space
tool-update-plan-arg-progress-schema-files-changed-item ::= "\"" char{0,1000} "\"" space

tool-update-plan-arg-steps-schema ::=
"[" space tool-update-plan-arg-steps-schema-item
("," space tool-update-plan-arg-steps-schema-item){0,127} "]" space

tool-update-plan-arg-steps-schema-item-goal ::= "\"" char{1,16384} "\"" space
tool-update-plan-arg-steps-schema-item-id ::= "\"" char{1,16384} "\"" space
```

After grammar compilation fails, llama.cpp may later report:

```text
common_chat_peg_parse: unparsed peg-native output
```

Disabling only `update_plan` makes the issue disappear in my setup.

Could the `update_plan` schema use smaller `maxLength` / `maxItems` limits, or otherwise provide a llama.cpp-compatible schema variant?

gaoanze888:
## Summary

Align Online Context Compact's runtime plan parser with the `update_plan` TypeBox schema, and keep that schema compilable by llama.cpp's native tool grammar.

The public schema previously allowed `id` and `goal` values up to 16,384 characters. Besides the original runtime mismatch, llama.cpp lowers those constraints to `char{1,16384}` and rejects the entire tool grammar because the repetition exceeds its 2,000-rule safety threshold (#65; upstream ggml-org/llama.cpp#25923).

This change:

- defines one TypeBox plan-string contract for both the registered schema and runtime parser;
- uses the Pi-loader-supported `typebox` and `typebox/value` public aliases;
- preserves TypeBox grapheme-cluster semantics for multilingual input;
- lowers explicit plan-string repetitions to a conservative 1,000, matching the existing progress-string limit;
- audits every registered `update_plan` `minLength`/`maxLength`/`minItems`/`maxItems` bound to prevent reintroducing an oversized grammar repetition;
- migrates previously valid 1,001–16,384-byte persisted plans by clearing only the active plan while preserving the latest request counters and cache-debt accounting.

The constant remains internal to the mechanism and is not added to the public barrel export.

Fixes #65.

## Validation

Exact head `db582d4` on current `main@2b79168`:

- `npm run check`
  - typecheck passed
  - 19 test files / 143 tests passed
  - package dry-run passed
- `node scripts/check-pi-compat.mjs`: passed
- real `DefaultResourceLoader` / `AgentSession` package integration: passed as part of the suite
- `git diff --check`: passed
- independent final review: no Blocker / High / Medium findings

The schema-threshold regression is local and deterministic; live verification against the reporter's llama.cpp backend remains external because this checkout has no configured llama.cpp provider.

AI-assisted: candidate discovery, implementation, and review used pi with GPT-5.6-sol. The final diff was independently reviewed.


gaoanze888:
Integration note after #63 merged: the new real package-loader test exposed one compatibility issue on this head. `plan.ts` imports `typebox/guard`, but Pi 0.85.1\u0027s `DefaultResourceLoader` aliases only `typebox`, `typebox/compile`, and `typebox/value`; jiti rewrites the subpath under the aliased `build/index.mjs` and extension loading fails with `Cannot find module .../typebox/build/index.mjs/guard`. I reproduced this by applying #59 onto current `main@2b79168`: the standalone plan tests pass, but `tests/pi-package-integration.test.ts` fails before extension registration.

A compatible equivalent is to use the loader-supported public imports:

```ts
import { Type } from "typebox";
import { Check } from "typebox/value";

const PLAN_STRING_SCHEMA = Type.String({ minLength: 1, maxLength: MAX_PLAN_STRING_LENGTH });
// ... Check(PLAN_STRING_SCHEMA, value)
```

I tested that form on the combined tree: both the Unicode boundary suite and the real package integration test pass (2 files / 7 tests). Please rebase onto current main and switch the guard import before merge; I can re-review the updated exact head promptly.

gaoanze888:
Updated and rebased onto current `main@2b79168` after the #63 integration test exposed the loader boundary. The runtime validator now uses the Pi-loader-supported public aliases `typebox` + `typebox/value` (`Type.String` + `Check`) while preserving TypeBox grapheme semantics and the shared max-length constant.

Validation on exact head `e8bd76e`:
- `npm run check`: 19 files / 142 tests, typecheck, package dry-run
- `node scripts/check-pi-compat.mjs`: passed
- real `DefaultResourceLoader` / `AgentSession` package integration: passed as part of the suite
- `git diff --check`: passed
- independent re-review: no Blocker / High / Medium

The branch is clean and now based on the merged Pi 0.85.1 update.

PR #60, opened 2026-09-15
fix(action-fusion): report savings only on success

gaoanze888:
## Summary

Only claim an Action Fusion saving after the fused command actually succeeds.

Previously, both `renderCall` and `renderResult` displayed a green `Money saved · 1 model round-trip avoided` banner whenever `then_run` was present. That also labeled pending calls, mutation failures, changed-file skips, command failures, aborts, and partial results as successful savings. The execution-side notification already used the correct success boundary.

This change:

- keeps pending edit/write calls on their built-in neutral renderer;
- recognizes only an exact `[then_run:succeeded]` result marker, optionally followed by command output;
- decorates only final, non-error results;
- preserves the built-in edit/write renderer for every state;
- adds the same outcome matrix for both edit and write, including embedded false markers, partial results, errors, failures, and skips.

## Validation

- `npm run check`
  - typecheck passed
  - 18 test files / 138 tests passed
  - package dry-run passed
- `git diff --check`

AI-assisted: candidate discovery, implementation, and review used pi with GPT-5.6-sol. The final diff received an independent review with no Blocker/High/Medium findings.


gaoanze888:
Rebased onto current `main@2b79168` after the Pi 0.85.1 update. Validation on exact head `b1173e4`: `npm run check` passed (19 files / 139 tests, typecheck, package dry-run), `node scripts/check-pi-compat.mjs` passed, the new real package-loader / AgentSession integration test passed, and `git diff --check` passed. No behavioral changes beyond the existing focused fix.

PR #61, opened 2026-09-15
fix(config): normalize reducer route identifiers

gaoanze888:
## Summary

Normalize surrounding whitespace in the configured Evidence-Preserving Reducer provider and model identifiers.

Both runtime configuration and the standalone preflight already validated these fields with `trim().length`, but returned the original string. A route such as `" openai-codex "` therefore passed preflight and startup, then failed the exact `modelRegistry.find(provider, model)` lookup only when a large eligible result triggered the reducer. The mechanism silently fell back despite the intended model being available.

This change returns the normalized identifier from both validators, keeping runtime and preflight behavior identical. Internal whitespace is left unchanged, and whitespace-only values remain invalid.

## Validation

- `npm run check`
  - typecheck passed
  - 18 test files / 141 tests passed
  - package dry-run passed
- runtime config regression covers spaces, tab, and newline
- preflight regression verifies normalized `effective_config`
- `git diff --check`

AI-assisted: candidate discovery, reproduction, implementation, and review used pi with GPT-5.6-sol. Final independent review found no Blocker/High/Medium findings.


gaoanze888:
Rebased onto current `main@2b79168` after the Pi 0.85.1 update. Validation on exact head `2415cdd`: `npm run check` passed (19 files / 142 tests, typecheck, package dry-run), `node scripts/check-pi-compat.mjs` passed, the new real package-loader / AgentSession integration test passed, and `git diff --check` passed. No behavioral changes beyond the existing focused fix.

lyttttt3333:
Independently validated exact head `2415cddfcfa57be946065fbd8cecd652932fece6`. LGTM; no blocking issues found in this validation.

- Reproduced the bug on baseline `2b79168`: surrounding spaces, tabs, newlines, mixed whitespace, and Unicode whitespace passed configuration checks but failed actual Pi model lookup, producing `reducer-model-unavailable` without a provider call. The clean-name control succeeded.
- On this PR, all those routes normalized consistently in preflight and runtime, resolved through the real Pi ModelRegistry, and completed EPR reduction with validated receipts. We checked the archived source/hash, retained failure evidence, and preservation of the original tool error status.
- Independent probes covered 17 scenarios per environment, including rejection of empty/whitespace-only/non-string values and preservation of internal whitespace.
- On both Pi 0.85.1 and the originally supported 0.84.2, `npm run check` passed: 19 test files / 142 tests, type checking, and package inspection. Public API checks and `npm audit --audit-level=high` also passed; the existing two moderate Vitest-related findings remain.

Validation used isolated dependencies and credential-free HOME directories. The 0.84.2 copy adjusted only dependency pins, its lockfile, and installation-guide version mentions; source and tests were unchanged. The independent probe exercised real configuration loading, feature registration, Pi model lookup, and EPR processing with deterministic local provider responses, not a live external API. No claims about live authentication or token savings.

lyttttt3333:
One clarification on scope and priority: this appears to be an edge case in manually authored or generated configuration, rather than a common issue in the normal configuration path. We have not measured how frequently users encounter it, so our validation should not be read as evidence of widespread impact.

The fix is small and reasonable configuration hardening, and our correctness assessment remains unchanged. The successful validation establishes that padded identifiers are handled correctly; it does not establish that this is a frequent or high-priority production failure.

HaozheLiu-ST (approved):
LGTM. Thanks @lyttttt3333 for the independent validation. The current head matches the version you checked. Combined validation with #9 and #35 also passes.

PR #62, opened 2026-09-15
fix(compaction): preserve cumulative cache debt

gaoanze888:
## Summary

- preserve unpaid cache-write debt across successive native compactions instead of replacing it with only the newest charge
- carry forward the per-request savings from earlier compactions when computing the combined breakeven horizon
- price each new cache-write charge from the estimated post-compaction context rather than the larger pre-compaction context
- document and test the debt ledger lifecycle, including unrelated native compactions and user corrections

## Why

Online Context Compact already carries outstanding debt into the next compaction decision, but a successful second compaction replaced both the outstanding debt and its repayment rate. For example, an existing `900` token debt repaid at `300` tokens/request followed by a new `800` token charge repaid at `200` tokens/request became `800 / 200`; the correct cumulative ledger is `1700 / 500`.

The new write charge also used the full pre-compaction context. A compaction that replaces `archiveTokens` with a `memoTokens` summary writes an estimated post-compaction context of:

```text
writeTokens - (archiveTokens - memoTokens)
```

Charging the removed prefix again overstated the cache-write premium and could defer compactions whose real breakeven fits the remaining request horizon.

This change keeps the ledger internally consistent:

- provider request: subtract the cumulative per-request savings
- successive SoL-Pi compaction: add its post-compaction write debt and savings to the outstanding ledger
- unrelated native compaction: preserve the ledger with zero new measured charge
- user correction: clear debt and request-horizon history

The calculation remains an estimate: Pi may retain more than the configured tail at message boundaries, and the actual summary can differ from the 1,000-token estimate. Decision and persisted debt now use the same estimate.

## Validation

- `npm run check`
  - TypeScript passed
  - 19 test files / 145 tests passed
  - package dry-run passed
- focused Online Context Compact suites: 3 files / 21 tests passed
- `node scripts/check-pi-compat.mjs`
- `git diff --check`
- two independent reviews validated the post-compaction accounting formula; one noted the existing summary/tail estimation uncertainty documented above


gaoanze888:
Rebased onto current `main@2b79168` after the Pi 0.85.1 update. Validation on exact head `656534a`: `npm run check` passed (19 files / 144 tests, typecheck, package dry-run), `node scripts/check-pi-compat.mjs` passed, the new real package-loader / AgentSession integration test passed, and `git diff --check` passed. The debt-ledger behavior is unchanged.

gaoanze888:
Additional combined validation: starting from `main@2b79168`, I applied #59, #60, #61, and #62 in order. All patches applied cleanly and `npm run check` passed with 19 files / 147 tests, including the real package-loader / AgentSession integration; Pi compatibility and diff-check also passed. The four PRs remain independently reviewable.

